PR #109, opened 2012-09-22
Add caseInsensitive option to parser builder

### LINKED ISSUE #108: Add a global option to make generated parser case insensitive
If the parser builder could accept an option like `{ caseInsensitive: true }`, and all parsing expressions will be case-insensitive, whether they have been appended with `i` or not, that would be great.

curvedmark:
Fix #108


dmajda:
Sorry I didn’t get to look at this, closing as obsolete now. See also my [comment](https://github.com/pegjs/pegjs/issues/108#issuecomment-128674289) at #108.

Issue #121, opened 2012-11-13
Trace option


ganeshv:
TLDR: Patch for tracing rules, helps greatly with debugging of new grammars.

While pegjs is way easier on the brain than BNF-based grammars, debugging a new grammar can be difficult. The generated code is so straightforward and simple that one can easily insert debug prints in rule-matching functions, but that gets wiped the moment you change a rule and regenerate the parser.

So I wrote a small patch which will print a trace of all the rules, the state of the input string at the point the rule is entered, and what was matched at leaving a rule.

To use it, you must invoke pegjs with the --trace command line option. e.g.  
`pegjs --export-var parser --track-line-and-column --trace foo.pegjs`

Additionally, you need to add a few options while invoking the parser:

```
var ast = parser.parse(source, {'startRule': 'start', 'trace': ['BlockList','Function',
'StatementBlock',  'IfStatement', 'EmptyStatement', 'EOS', 'Command' ,
'StringLiteral', 'Concat'], 'tabspace': 2, maxrules: source.length * 100});
```
-  The `trace` option is a list of rules for which trace output is emitted. This is to enable you to quickly narrow down the problem rule without wading through a ton of spam from very low-level tokenizer rules.
-  The `tabspace` option specifies how many spaces to indent each new rule.
-  The `maxrules` option is a half-assed safety net to catch infinite loops.

Sample output:

```
 Enter:BlockList:1:1:ls -l | gr
   Enter:Function:1:1:ls -l | gr
   Leave:Function:null
   Enter:StatementBlock:1:1:ls -l | gr
   ....
```

We entered BlockList, input line 1, column 1. Contents of input string (next 10 chars) "ls -l | gr"
BlockList = Function / StatementBlock. As you can see, the Function rule didn't match, so we try StatementBlock..

```
...
               Leave:SimpleCommand:{"ARGLIST":["ls","-l"]}
               Enter:SimpleCommand:1:9:grep foo
```

SimpleCommand matched and returned {"ARGLIST"..}. Now we try another SimpleCommand, input position is line 1, column 9, next few characters are "grep foo"...

Let me know if this sounds interesting. Code is lying in my fork, "trace" branch.

Issue #9, opened 2010-09-20
Other useful debugging output


celer:
So one thing that would be really awesome is some more debugging output. When I have a parser that doesn't work and is already producing a structured output, I am left either having to gut it to get back to something that lets me look at the structured output and see where the error is in my grammar. 

So it would be nice if there were some command line options for the parser generator, which caused the parser to dump debugging info, such as which rule fired, and what characters were captured as part of the rule.

hegemonic:
I would _love_ to have a `--trace` option. I'm currently trying to figure out where a parser is failing, and it's been an awfully slow process. With a trace option, I could have found the answer very quickly.

christofmuc:
+1 here. I totally totally love PEG.js and have used it for more than once thing already, but debugging a grammar is pita. Will try patching with the  suggestion above.

dmajda:
In the past few weeks I worked on this issue and added tracing support to PEG.js-generated parsers. Instead of implementing a simple static solution, I added a flexible API which allows users to pass their own *tracer* to a parser. This tracer receives events about parsing progress and can decide what to do with them (display, store, etc.). There is also a built-in default tracer (used when no tracer is supplied by the user) that logs events to the console in a simple format.

Details can be found in the [commit introducing tracing support](https://github.com/pegjs/pegjs/commit/da57118a43a904f753d44d407994cf0b36358adc).

The API is currently somewhat experimental (which is why it isn’t documented properly yet) and I expect it will evolve over time as experience is gained.

The default tracer is also somewhat bare-bones. I hope that PEG.js user community will develop more sophisticated tracers over time and I’ll be able to integrate their best ideas into the default tracer.

Big thanks goes to @ganeshv who [implemented](https://github.com/pegjs/pegjs/issues/121) initial tracing support for PEG.js which I used as a source of inspiration.

With that, I am closing this issue. Request for improving the API or the default tracer should go into separate issues.

PR #123, opened 2012-11-13
Added trace option for debugging grammar.

ganeshv:
See issue #121 for details.

I removed the iffy maxrules code, so this commit contains only the trace functionality. Very tiny, and negligible impact if you don't enable the trace option.


dmajda:
@ganeshv I ended up implementing tracing support in #9 in somewhat [different way](https://github.com/pegjs/pegjs/issues/9#issuecomment-87656481), but I want to give you a big thanks — your code and design you described at #121 served as an inspiration for me and it influenced the design I ultimately used a lot.

PR #130, opened 2012-12-03
Added trace (based on ganeshv's pull request) to master branch

wwoods:
Pretty much the same as ganeshv's, but against master instead of 0.7.0 proper (aka, uses line() and column()).  Also fixed the brace issue michaelficarra commented on, at least I think.


dmajda:
@wwoods I ended up implementing tracing support in #9 in somewhat [different way](https://github.com/pegjs/pegjs/issues/9#issuecomment-87656481). Sorry I didn’t look into this earlier, but I focused at other things at the time. Closing.

PR #180, opened 2013-06-06
Allow labels to propagate though sub-expressions

asterick:
That is an example of doing lua style comments in peg.js, which is currently impossible without a lot of creative work.


dmajda:
@asterick Since the tests I requested never materialized, I [re-implemented this myself](https://github.com/pegjs/pegjs/commit/fb5f6c6ee94b962c45f591f64b293bc11ba57ae6) — giving you credit in the commit message of course.

Closing.

PR #206, opened 2013-08-26
Makefile tweaks to support packaging

dpocock:
Hi,

I'm going to package your project, PEG.js, for Debian

I needed to make a couple of small tweaks to the Makefile, this allows us to create a source tarball for distribution on the Debian source DVD.  That means somebody without a network connection will be able to build it completely offline.

Regards,

Daniel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,9 @@ VERSION_FILE = VERSION
 
 # ===== Executables =====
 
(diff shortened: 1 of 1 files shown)


dmajda:
Closing this PR which got stalled (my fault).

If there is interest in reviving this, I have no objections against the `dist` target and against including non-minified version of JavaScript libraries used in the benchmark.

PR #208, opened 2013-09-09
Improve report-left-recursion performance, do not create a new array copy at each step

palant:
`report-left-recursion.js` compile step can require excessive amounts of time for some grammars. Almost 50% of that time is spent copying arrays around, fixed by this change.


dmajda:
I ended up [solving this slightly differently](https://github.com/pegjs/pegjs/commit/cdeecf750f8a3f781459ed817815a537392993c1), closing.

Issue #128, opened 2012-12-02
Dynamically get rule name from within an action


roded:
It would be useful to be able to retrieve the current rule name for use inside actions.
As an example for usage: building an AST from parsed text could utilize the current rule name as the node's type - without requiring specifying  the rule name explicitly.
Schematic example:

``` javascript
{
    // having a ruleName() would simplify grammar maintenance  
    var tok = new token("tokenType", offset);
    tok.addChild(a);
    tok.addChild(b);
    return tok;
}
```

dmajda:
I moved this after 1.0.0 as I’d like to get to that version quickly. I’m now mostly focusing at polishing existing functionality and I’m deferring minor features that “only” fix annoyances.

Also, the implementation of this is somewhat harder than one might think — it has to account for inlined rules ([currently only proxy rules](https://github.com/pegjs/pegjs/blob/d7fc0b5c3bfbcd079712d2d4958ec38eb9ad3822/lib/compiler/passes/remove-proxy-rules.js), but this will change in the future) and ideally have no performance impact.

balagge:
This is an old issue, but I would still be interested. 
Even if inline rules are not supported (e.g. the rule name is undefined for those)

StoneCypher:
This would be *extremely* useful to me, and if @dmajda will help me get this repo back into a healthy, releasable state, by re-cutting `0.10.0` as `0.12.0`, then I will happily make an effort to include this, in a fashion that resembles `.text()`

This would be one of the very early features I'd like to add

dmajda:
> This would be _extremely_ useful to me, and if @dmajda will help me get this repo back into a healthy, releasable state, by re-cutting `0.10.0` as `0.12.0`, then I will happily make an effort to include this, in a fashion that resembles `.text()`

I no longer have any rights to the project. Resolving this is up to @futagoza. Thanks for understanding.

StoneCypher:
@dmajda - oh :(  i didn't realize that.

ok.  i hope you might consider encouraging him to release this work back to the community, since he's more likely to listen to you than anyone else

that said, i won't bother you about this again.  i'm sorry.  thank you for one of the most important libraries in my life.  it's been wonderful for me.

----

@futagoza - we need you to please consider allowing someone else to maintain this project now.  nothing has been released or maintained since you took over, and you said that you're in the process of throwing everything away

[...]

PR #219, opened 2013-11-07
Added ruleName() function which can be called inside semantic actions to return the currently executing rule

EricMCornelius:
Saw mention of this as a desired feature in the issues list, and found myself needing it as well.  Not sure whether this is the most desired approach vs. storing a separate string table and using a numeric type for rule_id lookup.


dmajda:
This PR got stalled — sorry about that. I plan to look at this feature sometime after 1.0.0 (see #128), closing now.

Issue #30, opened 2011-08-11
Ability to specify repetition count (like in regexps)


ghost:
It would be helpful if the PEG.js grammar allowed something like range expressions of POSIX basic regular expressions to be used. E.g.:
- ```
  "a"\{1,7\}
  ```
  
   matches `a`, `aa`, ..., `aaaaaaa`
- ```
  "a"\{0,1\}
  ```
  
   matches the empty string and `a`
- ```
  "a"\{,6\}
  ```
  
   matches a string with up to (and including) six `a`'s
- ```
  "a"\{6,\}
  ```
  
   matches a string of six or more `a`'s
- ```
  "a"\{3\}
  ```
  
   matches only `aaa`, being equivalent to `"a"\{3,3\}`

signalwerk:
Hello. I have a complex file format to parse. It is half binary, half ASCII. 

Here a simplified version of the problem:

`KK4TesRandom` or `KK10TestATestBRandom`

The logic:

`<StringIndicator><StringLength><String><otherStuff>`

The `KK` is the indicator to mark a string. The following digits (here `4` and `10`) are the length of the string. Then the string itself (here `Test` and `TestATestB`). The string ist not terminated by any predictable pattern. I basically have to use the length information. I'd say this is a common pattern in binary file formats but is it possible to parse with the current grammar? 

Thank you.

Mingun:
I implement such thing in my branch [ranges-dynamic-boundary](https://github.com/Mingun/pegjs/tree/ranges-dynamic-boundary). Grammar will look so:
```
start = len:nx data:.|len| { return data; };
nx = n:$[0-9]+ { return parseInt(n, 10); };
```

signalwerk:
@Mingun wow! That works like a charm! Thanks a lot for your implementation and the short example. I made a few tests and it works awesome. I hope the your pull-request gets accepted to the master.

michael-brade:
I would love repetition counts as well. But I would suggest a slightly different syntax. [Pegasus](https://github.com/otac0n/Pegasus) is almost identical to pegjs, only for C#. See here: https://github.com/otac0n/Pegasus/wiki/Syntax-Guide#expressions

And they implemented this feature using this: `d<3> e<2,> f<1,5>`

evanpurkhiser:
What are peoples work arounds for this? I'm just getting into PEGjs right now, so maybe I'm trying to turn a screw with a hammer, but I'm just trying to match between 1 and 6 digits :)

[...]

PR #225, opened 2014-01-04
Improve error messages for GrammarError

Mingun:
`report-missing-rules` and `report-left-recursion` now has info, where problem found (line and column).

For example:
```
start
  = missed;
```
Now, error message `Referenced rule "missed" does not exist.`. After merging pull request it will be `Line 2, column 5: Referenced rule "missed" does not exist.`

```
start
  = b;
b = start;
```
Now, error message `Left recursion detected for rule "start".`. After merging pull request it will be `Line 3, column 5: Left recursion detected for rule "start".`


dmajda:
For the record, I implemented this somewhat differently in a [recent series of commits](https://github.com/pegjs/pegjs/compare/e75f21dc8f0e66b3d87c4c19b3fcb8f89d9c3acd...eaca5f0acf97b66ef141fed84aa95d4e72e33757).

@Mingun Here is a short feedback, as [promised](https://github.com/pegjs/pegjs/issues/246#issuecomment-89727941) earlier:

The way I see it, this PR solved too little and simultaneously too much.

It solved too little because it added a `region` function to the parser API without any consideration about naming (no discussion why the name `region`, not any other commonly used names for this kind of data) and the API (what location/region information should/should not contain). There was no attempt to unify the location reporting in wider scope (e.g. with syntax errors). There were no tests and no documentation.

It solved too much because it touched both the problem of the location information format (+ related APIs) *and* the problem of adding location information to places where it wasn’t available before. I see these two as distinct things.

Coupled with this were questionable choices in the API (a boolean parameter — something to be avoided unless the meaning will be obvious at call site), in the implementation (embedding location information into messages in `GrammarError` — especially while keeping it out in `SyntaxError`s in generated parsers) and quality problems (non-adherence to used code style and naming). 

Based on all this I decided my time would be better spent on doing my own implementation than working with you on putting this one into acceptable shape.

Beside changes obvious from my comments above probably the biggest change I did was in the name: I used `location` instead of `region`. I think this name describes stored information better and similar names and concepts are used in other tools/standards (e.g. [Mozilla JavaScript Parser API](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API), [Nitra](https://confluence.jetbrains.com/display/Nitra/Home), etc.). I also chose not to include the text (it won’t be needed in most cases and can be easily constructed from the location information and the input).

Although I didn’t directly use your code, your PR was still useful as inspiration and an indicator of demand. Thanks!

And one final note: I highly recommend having a look at Greg Kroah-Hartman’s presentation [I don’t want your code](https://github.com/gregkh/presentation-linux-maintainer/blob/master/maintainer.pdf) ([transcript](https://github.com/gregkh/presentation-linux-maintainer/blob/master/maintainer.txt)). Maybe it will make you see things from the other side a bit :-)

Mingun:
> It solved too little because it added a `region` function to the parser API without any consideration about naming (no discussion why the name `region`, not any other commonly used names for this kind of data) and the API (what location/region information should/should not contain).

Simply from my point of view not English-speaking the `region` was the best option: `location`, `position` usually implies one point in data whereas gives the name selected by me to association with a segment.

> There was no attempt to unify the location reporting in wider scope (e.g. with syntax errors).

Yes, I didn't think that it was possible and to thrust there. But this problem was easily solved by review a code. Nevertheless anybody within a year didn't decide to specify it :)

> There were no tests and no documentation.

Well, in this aspect my PR differs in nothing from your implementation as I see :)

> It solved too much because it touched both the problem of the location information format (+ related APIs) _and_ the problem of adding location information to places where it wasn’t available before. I see these two as distinct things.

The strange at you approach. Tests and the master code on me, so much more different things, than it. It not condemnation, is simple to me really unclear.

> Based on all this I decided my time would be better spent on doing my own implementation than working with you on putting this one into acceptable shape.

Anything terrible. I too like everything to keep under control and the slightest deviation from set by me standards is perceived by the unacceptable. It is simpler to me to make most as I consider it necessary, than to convince someone. So everything is normal. But only in this case I try to give at least feedback what exactly isn't pleasant or to do me the implementation in acceptable time limits (two years for such trifling thing as the range -- on me, it too. I would understand if the ton of work was required, but really it is impossible to find half an hour in two years to make review or the own implementation?)

> Although I didn’t directly use your code, your PR was still useful as inspiration and an indicator of demand. Thanks!

I am glad that it appeared the useful. Now I request implementation of #30. It is absolutely trivial, really :). Also, over it it is possible to realize dynamic boundaries (see https://github.com/Mingun/pegjs/tree/ranges-dynamic-boundary), which my be useful for many binary protocols.

Issue #234, opened 2014-01-29
Have a way to allow starting the parsing from any rule


krisnye:
Or else allowing any start rule should be the default if that property isn't specified.

dmajda:
Hmm, `*` as a special value has two unfortunate properties:

  * It needs to be escaped on the command-line.
  * It may lead to an impression that it is possible to specify allowed rules using globs.

A special string value (like `allowedStartRules = "all"` I proposed above) is fine in the API, but could be confusing in the CLI (does `--allowed-start-rules all` mean “all rules” or “the rule named `all`”?).

I’m not sure what the right solution is right now. I’m also starting to doubt this feature would be really that useful. The unit tests use case can be solved without this by parsing whole inputs, not just parts, in the tests — and this is probably the right way, because you generally don’t want to test pieces of a parser or a parser generated differently than the production one (where you would presumably not allow starting from any rule).

In any case, I don’t want to sped time on this now ⇒ moving to post-1.0.0. Comments describing real-world situations where this feature would enable/simplify something are welcome.

jamestalmage:
> I’m also starting to doubt this feature would be really that useful. The unit tests use case can be solved without this by parsing whole inputs, not just parts, in the tests 

I disagree. Testing an entire input makes it difficult to troubleshoot rules that are deeply nested, and moves you closer to integration tests vs unit tests.

I solved this slightly differently. I used a RegExp to parse my grammar for rule names, and added helper functions.

```js
var ruleNames = extractAllRuleNames(grammarSource);
// returns an array of rule names: ["RULE1", "RULE2"];

var parser = PEG.buildParser(grammarSource, {allowedStartRules: ruleNames});

// for each rule in ruleNames, generate a helper function that matches the rule name
parser.RULE1 = function(input, options) {
  options = options || {};
  options.startRule = "RULE1";
  parser.parse(input, options);
};
parser.RULE2 = function(input, options) {
  options = options || {};
  options.startRule = "RULE2";
  parser.parse(input, options);
};
// ... etc
```

This has the advantage of allowing you to explicitly specify which rule you expect to be the start rule. I think that is a bit better than just allowing it to start with any rule.

The only downside to this approach is that the regular expression that backs `extractAllRuleNames` is fairly brittle (especially when trying to handle the initializer portion). It would be much nicer to have all this automatically exported. Perhaps just by using a `--debug` flag when generating the parser. None of this makes much sense to me outside of testing.

felipellrocha:
[...]

PR #253, opened 2014-04-21
Allow any defined rule as an allowed start rule

andreineculau:
If allowedStartRule is the wildcard (i.e. "*"), then any defined rule
can be used as as startRule for the parser.

Fixes https://github.com/dmajda/pegjs/issues/234


dmajda:
Thanks for the patch, but I’m not sure about `*` and I’ve actually started to doubt sensibility the whole feature. See my [comment](https://github.com/pegjs/pegjs/issues/234#issuecomment-128943224) in #234 for details. Closing since I won’t be merging any time soon.

PR #257, opened 2014-04-28
Add CommonJS and AMD style module support

vvakame:
I need CommonJS module support and Web browser support by 1 source.


dmajda:
@michel-kraemer I agree that the current `Makefile` way is a bit clumsy and I’m not against replacing it with something better.

Being a Unix guy who likes how `make` works in general and who thinks that CLI is a reasonable way how to integrate build tools, I’m a bit cautious about Grunt, Gulp, etc. They partially reinvent the wheel, require plugins (= new dependencies), and it feels like they may not be the end of JavaScript build tool evolution (= possible rewrite in few months). But at this point I’m willing to give Gulp a try — it’s not like we’re building a huge project where possibility of future build system rewrite would be a big problem.

As for AMD modules, I was a bit cautious about them too (feeling they may be a temporary thing). But it seems they stuck and it makes sense to support them via UMD for the benefit of PEG.js users. Note this means two things:

  * Making PEG.js itself UMD module.
  * Making generated parsers UMD modules.

I think it makes sense to address both at the same time.

If you’d like to tackle this, I’d really appreciate it and I’d happily integrate your work.

andreineculau:
@dmajda "but at this point" - sounds like something changed, but it's not clear what. How is now more "make alternatives" than the past?

If the Makefile seems clumsy, then that's something to be fixed alone. If PEGjs and its generated parsers need to be UMD-ified then that's also something to be fixed alone.

I am biased as I'm also into keeping it simple with Make, but I think I'm still rather objective when saying that this and #312

1. fixes 2 things in one "gulp" (pun intended)
2. removes a "dependency" on make (everyone has make actually, except Window devs - so it's actually a dependency on cygwin) for PEGjs developers/building platform, not users/running platform (I think it's an important distinction)
3. introduces real dependencies instead - gulp + plugins
4. ports the "clumsyness" from the current Makefile to gulp tasks

Regardless of my take, thank you for the PR, Michel!

dmajda:
> @dmajda "but at this point" - sounds like something changed, but it's not clear what. How is now more "make alternatives" than the past?

Gulp got popular enough, stable enough, and there is a big ecosystem of tools around it. While all I wrote above still holds, the balance slowly shifted to a point where I’m willing to accept an external patch to replace `Makefile` by `gulpfile.js`.

> If the Makefile seems clumsy, then that's something to be fixed alone. If PEGjs and its generated parsers need to be UMD-ified then that's also something to be fixed alone.

Yes, but the latter may be a bit easier once the former is done.

> I am biased as I'm also into keeping it simple with Make, but I think I'm still rather objective when saying that this and #312
> 
> 1. fixes 2 things in one "gulp" (pun intended)

No, it changes only the build system.

> 2. removes a "dependency" on make (everyone has make actually, except Window devs - so it's actually a dependency on cygwin) for PEGjs developers/building platform, not users/running platform (I think it's an important distinction)

Agreed. Still, I see this as quite important. 

> 3. introduces real dependencies instead - gulp + plugins

Agreed.

> 4. ports the "clumsyness" from the current Makefile to gulp tasks

To some degree, yes. But I think it will be easier to get rid of some of it in JavaScript code than in a `Makefile` plus maybe some scripts.

I think of this as a bit of an experiment. If Gulp turns out to be a headache, I’ll be the first one to revert back.

johnsoftek:
One way to support npm modules in the browser is to use bundling tools such as browserify, webpack or jspm. There is a lot of momentum for this approach.

Standard module support will arrive with ES6 this year - you can use transpilers like babel now. I guess that node will eventually support the ES6 module syntax.

I don't see a future for AMD.

dmajda:
This PR got stalled and side-tracked (my fault) plus it doesn’t solve the problem completely, so I’ll close it now. However, the idea of AMD support is valid and I created #362 to track it.

PR #264, opened 2014-05-17
Second attempt for ranges (#30) -- machinery

Mingun:
Implement all machinery for ranges (issue #30). Doesn't implement syntax -- for this purpose there will be separate pull request. As the result -- doesn't contain tests, they will be together with syntax. However, all tests are passed.

Introduce two new opcodes:
* IF_ARRLEN_MIN min, then part length, else part length
* IF_ARRLEN_MAX max, then part length, else part length

Introduce new AST node -- `range`, that contains expression and the minimum and maximum number of its repetition.
If `node.min` is `null` or isn't positive -- check of the minimum length isn't made.
If `node.max` is `null`, check of the maximum length isn't made.

Unlike previous pull request (#209), there are:
* no logic for reduce range expression to `optional`, `zero or more` and `one or more` expressions -- this must be made by special compiler pass;
* lengths are hardcoded in bytecode, instead of are read from a stack;
* no support for delimiters -- they will be in next pull request.


dmajda:

PR #265, opened 2014-05-17
Ranges: proposed syntax

Mingun:
Based on #264, contains syntax for defining ranges and tests. Syntax for ranges:
```
expression|exact|
expression|..|
expression|min..|
expression|..max|
expression|min..max|
```


dmajda:

PR #266, opened 2014-05-17
Ranges: support for delimiter between range elements

Mingun:
Contains only implementation, based on #264. Proposed syntax and tests in next pull request.


dmajda:

PR #267, opened 2014-05-17
Ranges delimiter proposed syntax

Mingun:
Implement proposed syntax for ranges with delimiters:
```
expression|  exact , delimiter|
expression|   ..   , delimiter|
expression|min..   , delimiter|
expression|   ..max, delimiter|
expression|min..max, delimiter|
```
That's mean, that the `delimiter` must encounter between `expression` elements in parsed input, but not before first, or after last element in sequence. Result -- array of `expression` elements, `delimiter` parts will be dropped.


dmajda:

PR #269, opened 2014-05-20
Appease jshint warnings

arlolra:
It claims "a constructor name should start with an uppercase letter" and is concerned about a "possible strict violation" in the use of `this`.


dmajda:
Re `newcap`: Sorry, I don’t think it makes sense to change conventions used in generated parser code just to appease JSHint. Moreover, as I just learned, [the `newcap` option is deprecated](http://jshint.com/docs/options/#newcap) and on its way out.

Re `strict`: No problem with this one in principle. Of course, the code needs to be converted first, which is tracked by #323 and #324.

Given that I won’t accept the `newcap` change I’m closing this. Any strict mode work should go into separate pull request(s).

And sorry I got back to this so late.

Issue #270, opened 2014-05-24
Report duplicate labels as errors


lukas-vlcek:
Currently (peg.js v0.8.0) it is possible to have more labels of the same name in the expression.

Trivial example of valid grammar:

    start = s:name s:suffix { return s+s; }

    name = "peg"

    suffix = ".js"

In case of complex grammars it can be hard to spot this situation visually and it can lead to issues which can take some time to nail down. It would be helpful to enable to report this as an error (if not strictly required it might make sense to at least add a new option to `PEG.buildParser`).

Relevant post of forum: https://groups.google.com/forum/#!topic/pegjs/O7HR-vBMEwg

dmajda:
@mulderr Thanks for tackling this issue and sorry I get back to you so late. Would you still be interested in finishing your code and getting it merged into PEG.js? If so, please let me know and create a pull request so that I can review.

andreineculau:
> I agree it makes sense to report duplicate labels as errors.

Not me. Reporting them as warnings, yes. Or as errors, with the ability to turn off the check, maybe.

But reporting them as errors simply removes a nice patch feature in pegjs `cat basic.pegjs additions.pegjs > amended.pegjs` and it just works. If they are reported as errors, Now you'd have to parse the grammars, find which rules already exist in basic.pegjs, remove them, then compile.

dmajda:
@andreineculau This issue is about labels, not rules. Concatenating grammars will still work even when this one is fixed.

That said, I think duplicate rules should be also an error, because redefinition of a rule is more likely an error than not, and your concatenating use case is not relevant for most users. I filed #318 to track this.

andreineculau:
Darn! I should've paid more attention and/or keep my mouth shut :)

mulderr:
@dmajda The pull request is here: https://github.com/pegjs/pegjs/pull/278

I have since abandoned my project which was using pegjs - reasons having nothing to do with pegjs itself. Anyway if some minor modifications would be needed to merge this I think I could spare some time to see it through.

futagoza:
@mulderr and @dmajda: I have fixed the issue in #322

PR #278, opened 2014-06-22
(feature) require label names to be unique within a rule (issue #270)

mulderr:
Add a new compiler pass to check if label names within a rule are unique.
Report any duplicate labels as a GrammarError.


dmajda:
This got obsoleted by #322, closing.

> This is a breaking change and would require revving the major version number.

@otac0n No, in 0.x.y anything is allowed per SemVer rules.

Issue #273, opened 2014-05-30
parse async


andreineculau:
Based on https://groups.google.com/forum/#!topic/pegjs/dL6HbQAC9Y0
I was thinking about an .abort() helper
which implies making .buildParser and .parse async

dmajda:
I don’t plan implementing async/interruptible parsing for at least next few months. However, if someone more interested than me tackles that as a plugin I’ll be really happy. It would be a great way to explore the API and issues without doing experiments in PEG.js mainline.

With the current bytecode-based architecture, the plugin should need just to replace the `generateJavascritpt` pass with its own.

Closing this issue.

PR #281, opened 2014-07-21
Allow an option to specify a start offset

arlolra:
* `options.startOffset` starts the parsing at that position in the input.


dmajda:
Sorry, but this option is not worth its weight, you can always do this instead:

```javascript
parser.parse(input.substring(startOffset));
```

Closing.

arlolra:
@andreineculau Yeah, we're using it to asynchronously parse top-level blocks of wikitext (w/ a fictious EOF as you said). We're using the start offset to preserve the backtracking cache to avoid some pathological parses. Having this option makes that a little less hacky.

@dmajda Does the above explanation lend credence to its worthiness?

subbuss:
Hi @dmajda We use pegjs in the Parsoid project at the Wikimedia Foundation as a high-level tokenizer. And, on some wikipedias, we run into some fairly large pages running into 100s of KBs (> 1MB pages are not uncommon), and for parsing performance, we tokenize the page in chunks asynchronously so we can get network and other I/O requests going without waiting for the tokenization of the entire page to complete..

This requires us to be able to parse input source starting at non-zero offsets while also preserving the backtracking cache in pathological scenarios (we see this often enough on wikipedias because of errors in wikitext). 

See https://gerrit.wikimedia.org/r/#/c/145683/3..5/lib/mediawiki.tokenizer.peg.js for a diff of how this feature simplifies our code. Plus, one of our team members also reported performance impacts from the workarounds. This feature is important enough that we are running our code with a fork with arlolra's fixes. It would be great to see this folded into the mainline repository so we can continue to track mainstream fixes and improvements to pegjs.

dmajda:
@arlolra @subbuss I looked at your [tokenizer](https://git.wikimedia.org/blob/mediawiki%2Fservices%2Fparsoid.git/3df9b319977fc5a00d954592a9777f5739862f44/lib%2Fmediawiki.tokenizer.peg.js) and [grammar](https://git.wikimedia.org/blob/mediawiki%2Fservices%2Fparsoid.git/3df9b319977fc5a00d954592a9777f5739862f44/lib%2FpegTokenizer.pegjs.txt) briefly. AFAICS, the whole thing is basically a big hack around the fact that PEG.js-generated parsers currently can’t interrupt and resume parsing. I’m impressed by the fact that you pulled this off, but I also think there is much cleaner way to accomplish what you want: Write a PEG.js plugin replacing the `generateJavascript` compiler pass with one that generates an interruptible parser.

I’m pretty sure that with such plugin you’d be able to simplify the tokenizer significantly and get rid of some of your grammar hacks. It’s a modular and maintainable solution. Also given that interruptible parser requests come up periodically (e.g. #273), the plugin would likely be useful for wider PEG.js community and I can imagine merging it (or something based on it) into PEG.js mainline eventually.

Let me know if you decide go this way. I’d be happy to assist if you’ll need help.

arlolra:
Sounds like a great idea. I'll be working on this in the near future.

PR #282, opened 2014-07-28
fix warnings reported by jshint

k2s:
jshint reports minor warnings on generated parser code. still it makes troubles becasue I use jshint in git hook before commit.

thank you for this great tool

### DIFF
--- a/lib/compiler/passes/generate-javascript.js
+++ b/lib/compiler/passes/generate-javascript.js
@@ -924,10 +924,10 @@ function generateJavascript(ast, options) {
     '          expectedDescs[i] = expected[i].description;',
     '        }',
     '',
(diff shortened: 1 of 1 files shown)


dmajda:
Sorry, but I won’t accept this change:

1. In case of continuation lines I  prefer operators to be at the beginning, not at the end.
2. The parser code as generated by `PEG.buildParser` is an expression, not a statement. Appending a `;` at the end is wrong.

More generally, generated parser code can’t conform to any random JSHint settings its users use. The only guarantee is that it conforms to settings expressed in [`.jshintrc`](https://github.com/pegjs/pegjs/blob/master/.jshintrc). If that doesn’t suite you, I recommend excluding the parser source from your JSHint checks.

PR #296, opened 2014-10-09
Fixed missing forward slash escape in regexp

joshvoigts:


dmajda:
@joshvoigts Thanks for your contribution but `\` doesn’t have to be escaped in a character class (as already noted). Closing.

joshvoigts:
No problem, thanks for looking into this. The javascript engine I was using choked on this, but I was able to change it by hand.

PR #299, opened 2014-10-25
Revert "PEG.js grammar: Disallow empty sequences"

Mingun:
This reverts commit df154daafb9c6c952351493af02d3a55e0b05c59.

Conflicts:
	lib/compiler/passes/generate-bytecode.js
	lib/parser.js
	spec/parser.spec.js
	spec/unit/compiler/passes/generate-bytecode.spec.js
	src/parser.pegjs

I think, change from df154daafb9c6c952351493af02d3a55e0b05c59 creates more problems, than decides. For example, I have a grammar describing the protocol of the ATM which looks approximately so (I use Ranges extension from #267):
```
GS 'Group Separator' = '\x1D' {return '<GS>'};
n4 'n4' = n:$[0-9]| 4| {return parseInt(n,10);};
n5 'n5' = n:$[0-9]| 5| {return parseInt(n,10);};
n7 'n7' = n:$[0-9]| 7| {return parseInt(n,10);};

TerminalState 'Terminal State'
  = '1' ConfigInfo                   // Send configuration information
//...
  / '7' ExtendedSupplyCounter|.., GS|// Send supply counters (extended)
//...
  ;
ConfigInfo = ...;
ExtendedSupplyCounter
  = 'A' tsn:n4 tranCnt:n7 // Transaction
  / 'B' cardsCaptured:n5  // Card Reader
//...
// Though all fields are marked with the identifier, however documentation
// says that they will go in strictly certain order (as it is specified here) and
// if any data aren't, there will be GS
  /  // <-- Problem. What here shall be?
  ;
```
The grammar shall accept such sequence (field `B` is omitted, but GS for this still present) but after this commit it is impossible:
```
...7A...<GS><GS>C...<GS>D...
```
I suggest to rollback it since it creates unnecessary complication of a parser and isn't obvious at all.

### DIFF
--- a/lib/compiler/passes/generate-bytecode.js
+++ b/lib/compiler/passes/generate-bytecode.js
@@ -414,14 +414,14 @@ function generateBytecode(ast) {
         }
       }
 
-      return buildSequence(
+      return node.elements.length > 0 ? buildSequence(
         [op.PUSH_CURR_POS],
         buildElementsCode(node.elements, {
           sp:     context.sp + 1,
           env:    context.env,
           action: context.action
         })
-      );
+      ) : [op.PUSH_EMPTY_ARRAY];
     },
 
     labeled: function(node, context) {
--- a/lib/compiler/passes/report-left-recursion.js
+++ b/lib/compiler/passes/report-left-recursion.js
@@ -11,7 +11,9 @@ function reportLeftRecursion(ast) {
     },
 
     sequence: function(node, appliedRules) {
-      check(node.elements[0], appliedRules);
+      if (node.elements.length > 0) {
+        check(node.elements[0], appliedRules);
+      }
     },
 
     rule_ref: function(node, appliedRules) {
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -69,9 +69,11 @@ module.exports = (function() {
                 : expression;
             },
         peg$c9 = function(first, rest) {
-              return rest.length > 0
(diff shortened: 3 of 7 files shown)


dmajda:
I stand by my decision to remove empty sequences, see [my reply](https://github.com/pegjs/pegjs/commit/df154daafb9c6c952351493af02d3a55e0b05c59#commitcomment-10667420) to [your comment](https://github.com/pegjs/pegjs/commit/df154daafb9c6c952351493af02d3a55e0b05c59#commitcomment-8301000) for details.

Next time you can save effort by initiating a discussion first (*especially* for a change that directly contradicts my previous explicit decision).

Mingun:
> Next time you can save effort by initiating a discussion first (especially for a change that directly contradicts my previous explicit decision).

Unless PR in itself isn't discussion? The GitHub catchword -- _social_ coding. But I understood your reasons, I really didn't think of it, as well as of the bypass decision. And still, I think that code generation was changed in vain. Yes, the PEGjs parser won't be able to create AST with a node in which there will be an empty sequence, but the user is able to do it. Such AST shall be processed normally, but not fall with an error.

dmajda:
> Unless PR in itself isn't discussion?

Just a note — to me, PR is a request to include a piece of code into the codebase. If it is meant as a start of a discussion it needs to be stated explicitly. And in general, I prefer discussions in issues, because that separates discussion about a problem from discussion about a specific solution.

Issue #45, opened 2011-08-25
Implement parametrizable rules


ceymard:
It would be great to be able to parametrize rules with variables ;

```
string
   = '\"' parse_contents '\"' ->
   / '\'' parse_contents('\'') '\'' ->
   / '+' parse_contents('+') '+' -> /* sure why not :) */

parse_contents(terminator='\"')
    = ('\\' terminator / !terminator .)+ -> return stuff
```

tebbi:
I just wrote a fork that supplies an environment, accessible using the variable `env`: https://github.com/tebbi/pegjs
 This is the same as the `#STATE{}` object suggested above.
It is a quick hack, using a (package-)global variable, which is restored whenever a parsing function is left. The copying of `env` is accomplished with Object.create().

qwertie:
I have the impression that PEG.js doesn't support parameters of any kind on rules - which is surprising. This feature is very important to me.

What I need is simpler than the OP's request - the OP wants to modify the grammar itself depending on the parameter, but at a minimum I just need to pass an integer into a rule. Basically I want to translate an [LLLPG](http://ecsharp.net/lllpg) rule that looks like this (where `PrefixExpr` is a high-precedence expression such as a prefix expression like `-x`, or an identifier...):

	@[LL(1)]
	rule Expr(context::Precedence)::LNode @{
		{prec::Precedence;}
		e:PrefixExpr(context)
		greedy
		(	// Infix operator
			&{context.CanParse(prec=InfixPrecedenceOf(LT($LI)))}
			t:(NormalOp|BQString|Dot|Assignment)
			rhs:Expr(prec)
			{ ... }
		|	// Method_calls(with arguments), indexers[with indexes], generic!arguments
			&{context.CanParse(P.Primary)}
			e=FinishPrimaryExpr(e)
		|	// Suffix operator
			,,, 
		)*
		{return e;}
	};
	// Helper rule that parses one of the syntactically special primary expressions
	@[private] rule FinishPrimaryExpr(e::LNode)::LNode @{
	(	// call(function)
		"(" list:ExprList(ref endMarker) c:=")"
		{ ... }
		|	// ! operator (generic #of)
			"!" ...
		|	// Indexer / square brackets
			{var args = (new VList!LNode { e });}
			"[" args=ExprList(args) c:="]"
			{ ... }
		)
		{return e;}
	};

[...]

PR #301, opened 2014-11-06
Template rules (implement #45)

Mingun:
I made support for template rules. The rule can be parameterized by other rules / expressions. All references inside template rule with name of template parameter refer to parameter, superimposing visibility scope of the rule with a similar name.

At the moment parameters can't be templates, i.e. such construction is forbidden:
```
Apply<X, Y> = X<Y>;
```

Perhaps, further I will make its support.


dmajda:
Thanks for the work, but I’ll close this now. Templates are something I’d like to have a look at after 1.0.0.

PR #305, opened 2014-11-26
Better hide pegjs internals from user code -- isolate user code namespace

Mingun:
Also this refactoring need for include feature, on which I now working. It requires isolation of user code, because may be several initializers from several included grammars. Because some grammars need some internals of pegjs to work (for example, `peg$currPos`) there is good idea also merge #225 when you merge this.

Perfomance (-1.6%)
---------------
Measured with `benchmark/index.html` on Opera 12.17 build 1863.
Before (1000 times)
1. no cache, speed:
    Total	**68.14** kB	**93.81** ms	**726.29** kB/s
2. cache, speed:
    Total	**68.14** kB	**147.44** ms	**462.14** kB/s
3. no cache, size:
    Total	**68.14** kB	**740.66** ms	**91.99** kB/s
4. cache, size:
    Total	**68.14** kB	**506.97** ms	**134.40** kB/s

After (1000 times)
1. no cache, speed:
    Total	**68.14** kB	**95.46** ms	**713.78** kB/s (-1.72%)
2. cache, speed:
    Total	**68.14** kB	**149.20** ms	**456.67** kB/s (-1.18%)
3. no cache, size:
    Total	**68.14** kB	**753.35** ms	**90.44** kB/s (-1.68%)
4. cache, size:
    Total	**68.14** kB	**516.26** ms	**131.98** kB/s (-1.80%)


dmajda:
Sorry, I don’t think that decreasing the number of visible internal variables is worth increasing internal code complexity, integrity (= moving parsing-related code out of `peg$parse`), and decreasing parser performance.

Users know that `peg$`-prefixed variables are internal and it doesn’t matter much if there is just one or many visible. And some variables will always be present given that one can’t create a new, fully isolated scope in JavaScript.

Mingun:
Parser perfomance may be restored, if caching result of `peg$sandbox` invocation in separate variables, but I think that the JS engine shall cope with it itself.

Primary benefit of separation of the user code from the generated -- opportunity to avoid name conflicts in case of including one grammar in another that is one of possible options of implementation of import of grammar (#308).

> And some variables will always be present given that one can’t create a new, fully isolated scope in JavaScript.

Such purpose never also didn't stand -- of couse, some varialbes always will be present.

> integrity (= moving parsing-related code out of  peg$parse )

I think that user actions and predicates aren't an parser-related code, because it can be replaced without breaking parser structure, as those so it is logical to carry out them from a parser.

dmajda:
> Primary benefit of separation of the user code from the generated -- opportunity to avoid name conflicts in case of including one grammar in another that is one of possible options of implementation of import of grammar (#308).

Yes, but that isn’t something I’m interested *right now* and — unless I’m mistaken — is possible to accomplish in a simpler way (without moving the definitions out of `peg$parse` and without performance degradation). The way you implemented this and your commit messages suggested to me that better runtime isolation was at least partially a goal.

> > integrity (= moving parsing-related code out of  peg$parse )

> I think that user actions and predicates aren't an parser-related code, because it can be replaced without breaking parser structure, as those so it is logical to carry out them from a parser.

Anything that calls functions defined in `peg$parse` is by definition parsing-related.

Moreover, the philosophy of PEG.js is that user code is integral part of the grammar.

PR #308, opened 2014-12-17
Import feature: include imported files

Mingun:
This depends on some other my pull requests:
- #305 -- for opportunity to have own scope for the initializer for each of the included grammars

I implement import through include in this PR, but, as you can see in branch https://github.com/Mingun/pegjs/commits/import-feature-syntax, this is not necessary. This is only one backend for import, is also possible and even it is necessary to implement others.

Syntax for declare import and use imported rules:
```js
// Note, before initializer code block
#alias1 = 'any/string/for/example/path/to/file/for/include/backend';
#alias2 = 'any count of imported grammars supported';
{
// Initializer
// If imported grammars has own initializers, it will be executed before that
// in order of import
}
start
  = #alias1 // use default (now, first, but can be configurable) rule in imported grammar
  / #alias2:someRule // use specified rule in imported grammar
```

Include backend recursively include AST of all imported grammar in one. You can say, how it must search imported grammars. It is possible even to directly generate grammar AST, generally reading nothing from a filesystem. See `Resolver` (https://github.com/Mingun/pegjs/tree/include-pass/lib/compiler/resolver.js).

Cyclic dependencies handled by `Resolver` object -- it throws `GrammarError` when detected them.

I can also create PR containing only syntax of import clause (branch https://github.com/Mingun/pegjs/tree/import-feature-syntax) on me it seemed that without any backend it will be useless


Qix-:
You should rebase on top of HEAD and then squash. I'd love to test this, personally.

dmajda:

PR #312, opened 2015-01-25
Converted Makefile to gulpfile.js

michel-kraemer:
As discussed in https://github.com/pegjs/pegjs/pull/257#issuecomment-69318916 I converted the `Makefile` to a `gulpfile.js`. The gulp file appears to be a little bit longer but it improves the build process as follows:

* We may now simply wrap PEG.js into an UMD module as dicussed in https://github.com/pegjs/pegjs/pull/257 by just adding `gulp-umd` or something similar
* You may now run the build on systems where `make` is not available (e.g. Windows without Cygwin)
* Therefore, contributor friction is reduced as anyone may now build PEG.js on any system with Node.js installed by just calling

    npm install -g gulp
    npm install
    gulp

The gulp file produces the exact same result as the Makefile. The only difference is that the minified bundle looks slightly different since I upgraded uglify.js.

I hope you like the PR. Let me know if you have questions.

Cheers,
Michel


dmajda:
Thanks a lot! I don’t have time for a detailed look now but I’ll try to do a review by the end of this week.

dmajda:
So, I finally managed to go through the pull request. Sorry it took so long — I’ve been busy at work and I also looked at Gulp in detail as part of the review, which delayed it further. 

Your pull request is of very high quality — big thanks for that! I’ll now post my comments. Don’t be mislead by their sheer amount — most of them are trivial or stylistic in nature.

It would be great if you could go through my comments, adapt your code accordingly, and post the result as a new pull request (so my comments here are preserved).

dmajda:
This got stalled (no reaction to my review comments), closing.

PR #315, opened 2015-02-06
fix generated parsers not working in strict mode

vith:
I had written a PEG that was working in the online demo but not in my app. There were two problems.

The first error I got was `generate-bytecode.js:406 Uncaught ReferenceError: failedIndex is not defined`.

Took me a while to figure out what was going on, because there's no such symbol `failedIndex` in either the latest commit to master or the 0.8.0 tagged commit.

I had installed pegjs with `npm install --save-dev pegjs`, which reports the version as 0.8.0. However, the `failedIndex` error goes away if I install pegjs with `npm install --save-dev pegjs/pegjs`, which is also version 0.8.0. No idea where npm got the broken version of 0.8.0, maybe it has something to do with the repository being moved from one namespace to another on github. As you can see here https://www.npmjs.com/package/pegjs npm still lists the package source as github.com/dmajda/pegjs.

Anyway, with `pegjs/pegjs` the `failedIndex` error is gone, but then I have the problem fixed in this pull request: `parser.js:4838 Uncaught SyntaxError: Unexpected strict mode reserved word`

So what's happening here is the javascript generator wraps the parser initializer in `{` `}`. This is invalid javascript code; it looks like a bare object but contains a block of code rather than object members.

The generated parsers don't include a `'use strict';` statement, so this normally just works anyway I guess. But I'm using webpack which is bundling a generated parser together with other modules and apparently strict mode is somehow applying to the parser, which causes the browser to bail out instead of executing the contents of the `{` `}` block.

My patch just stops wrapping the initializer in `{` `}`. It appears to work, but I'm wondering if the initializer and other code blocks should actually be more isolated from the rest of the generated code. I see most of the items defined inside `function peg$parse` are prefixed with `peg$` to prevent collisions, but there are a few names that a user might try to use: `input`, `options`, `text`, `offset`, `line`, `column`, `expected`, `error`. In fact, if I make an initializer like `{ input = null; }`, the parser breaks.

### DIFF
--- a/lib/compiler/passes/generate-javascript.js
+++ b/lib/compiler/passes/generate-javascript.js
@@ -965,7 +965,7 @@ function generateJavascript(ast, options) {
   }
 
   if (ast.initializer) {
-    parts.push(indent4("{" + ast.initializer.code + "}"));
+    parts.push(indent2(ast.initializer.code));
     parts.push('');
   }
 
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -4835,54 +4835,54 @@ module.exports = (function() {
       return s0;
     }
 
-    {
-      var OPS_TO_PREFIXED_TYPES = {
-        "$": "text",
-        "&": "simple_and",
-        "!": "simple_not"
-      };
 
-      var OPS_TO_SUFFIXED_TYPES = {
-        "?": "optional",
-        "*": "zero_or_more",
-        "+": "one_or_more"
-      };
+    var OPS_TO_PREFIXED_TYPES = {
+      "$": "text",
+      "&": "simple_and",
+      "!": "simple_not"
+    };
 
-      var OPS_TO_SEMANTIC_PREDICATE_TYPES = {
-        "&": "semantic_and",
-        "!": "semantic_not"
-      };
+    var OPS_TO_SUFFIXED_TYPES = {
+      "?": "optional",
+      "*": "zero_or_more",
+      "+": "one_or_more"
+    };
 
-      function filterEmptyStrings(array) {
-        var result = [], i;
+    var OPS_TO_SEMANTIC_PREDICATE_TYPES = {
+      "&": "semantic_and",
+      "!": "semantic_not"
+    };
 
-        for (i = 0; i < array.length; i++) {
-          if (array[i] !== "") {
-            result.push(array[i]);
-          }
+    function filterEmptyStrings(array) {
+      var result = [], i;
+
+      for (i = 0; i < array.length; i++) {
+        if (array[i] !== "") {
+          result.push(array[i]);
         }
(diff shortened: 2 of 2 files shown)


Mingun:
Perhaps, then also to add the `strictMode` parameter to `options` to generate a parser with the `'use strict'` statement in the begin? Then this PR will look finished.

dmajda:
> No idea where npm got the broken version of 0.8.0

Not sure how you got the idea that PEG.js 0.8.0 package is broken in any way. I just moved the PEG.js repository since I released 0.8.0 and I didn’t bump the version in Git yet (will do that just before the 0.9.0 release).

> Anyway, with `pegjs/pegjs` the `failedIndex` error is gone, but then I have the problem fixed in this pull request: `parser.js:4838 Uncaught SyntaxError: Unexpected strict mode reserved word`
>
> So what's happening here is the javascript generator wraps the parser initializer in `{` `}`. This is invalid javascript code; it looks like a bare object but contains a block of code rather than object members.

The initializer block is [perfectly valid JavaScript code](http://www.ecma-international.org/ecma-262/5.1/#sec-12.1). Your error message suggests that the problem with your parser was *inside* its initializer code.

The reason for wrapping the initializer code into `{` and `}` is mostly aesthetical — the output matches what was in the grammar exactly and the initializer code is visually separated from the parser code itself.

> My patch just stops wrapping the initializer in `{` `}`. It appears to work, [...]

Based on what I know and on the information you provided, there should be no difference.
 
> I'm wondering if the initializer and other code blocks should actually be more isolated from the rest of the generated code. I see most of the items defined inside `function peg$parse` are prefixed with `peg$` to prevent collisions, but there are a few names that a user might try to use: `input`, `options`, `text`, `offset`, `line`, `column`, `expected`, `error`. In fact, if I make an initializer like `{ input = null; }`, the parser breaks.

The unprefixed identifiers are exactly the ones the parser code *should* see. The fact that the code can assign to some variables beside just reading them is unfortunate but can’t be easily prevented.

As for the isolation, I’d like to have more of it, but it’s not possible because of JavaScript lexical scoping and for performance reasons.

As for the strict mode, it’s definitely worthwhile to have both PEG.js and the parsers it generates in strict mode, but it’s not a priority for me ATM. I filed #323 and #324 to track this.

dmajda:
> Perhaps, then also to add the `strictMode` parameter to `options` to generate a parser with the `'use strict'` statement in the begin? Then this PR will look finished.

Enabling strict mode as an option doesn’t make much sense, there are almost no situations under which one would want to disable it. (The only one I can think of is when you have an existing parser with a heap of strict-mode-incompatible code inside the initializer and actions. But if you are in such situation, strict mode is probably the last of your problems.)

Issue #318, opened 2015-02-13
Report duplicate rules as errors


dmajda:
PEG.js currently allows duplicate rule definitions and applies “last one wins” semantics for these. It should report an error instead. The reason is that a duplicate rule definition is most likely unintended and a result of user error.

futagoza:
@dmajda: I have fixed the issue in #322

Issue #316, opened 2015-02-13
Make labels behave like block-scoped variables


dmajda:
Labels should basically work like block-scoped variables, where “blocks” are created implicitly by `/`, `&`, `!`, `?`, `*`, `+`, `$`, sequences, labels and actions. This means that:
  * Actions should see all labels defined at the top level of the preceding expression or sequence *and* labels defined on levels above them.
  * Predicates should see all labels defined at the top level of the sequence they are embedded in that precede the predicate *and* labels defined on levels above them.

For example, parser created from the following grammar throws `a is not defined` error today, but it will return `["a", "a"]` once this issue is fixed:

```
start = a:"a" ("b" "c" { return a; })
```

Split off from #69.

Related pull request: #180

andreineculau:
I should check it out myself, but does this mean that it works the same between different rules?

```
start= a:"a" b {return [a,b]}
b = b:"b"
```

dmajda:
Label scope is of course limited to expressions in one rule. Anything else would be madness (think recursion).

PR #322, opened 2015-02-22
Disallowed duplicate rule and label names

futagoza:
This commit adds 2 pass's to the compiler that disallow:
- duplicate rule names (issue #318)
- duplicate label names (issue #270)

The duplicate labels pass is based on #278 (by @mulderr) but fixed, as the original was assuming all choices within a rule shared labels, which they don't. The spec test is updated as #316 removed the need for 2 checks within the *"reports duplicate labels inside expressions"* test. Also updated pass to allow the user to disable the check by setting the 'option.reportDuplicateLabels' to a false value ('false' or '0').

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,21 +5,22 @@ PEGJS_VERSION = `cat $(VERSION_FILE)`
 # ===== Modules =====
 
 # Order matters -- dependencies must be listed before modules dependent on them.
-MODULES = utils/arrays                          \
-          utils/objects                         \
-          utils/classes                         \
-          grammar-error                         \
-          parser                                \
-          compiler/asts                         \
-          compiler/visitor                      \
-          compiler/opcodes                      \
-          compiler/javascript                   \
-          compiler/passes/generate-bytecode     \
-          compiler/passes/generate-javascript   \
-          compiler/passes/remove-proxy-rules    \
-          compiler/passes/report-left-recursion \
-          compiler/passes/report-missing-rules  \
-          compiler                              \
+MODULES = utils/arrays                            \
+          utils/objects                           \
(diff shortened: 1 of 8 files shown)


mulderr:
Nice work and good catch with the choices thing.

@dmajda, this probably takes care of the modifications you had in mind in #270? Looks much better than my original pull request, so I would recommend going with this instead.

dmajda:
@futagoza If you’re still interested in this, I think the best way forward is to split this PR into two, one for each check. The `reportDuplicateRules` pass is OK and I’d be happy to take it in once small stylistic issues I noted are fixed. The `reportDuplicateLabels` pass needs more work.

futagoza:
[#commitcomment-10544235](#commitcomment-10544235) I don't mind removing this, but the only reason I implemented this option was because most import/include/concat methods people use often lead to duplicate rules, often intended. If this option isn't implemented they will have to modify their build systems to accommodate the requirement of unique rule names.

[#commitcomment-10544484](#commitcomment-10544484) I thought going through the `choice` rule fixed this issue, damn. I've checked how `context.env` is used in the [generateBytecode pass](https://github.com/pegjs/pegjs/blob/d7fc0b5c3bfbcd079712d2d4958ec38eb9ad3822/lib/compiler/passes/generate-bytecode.js), but following this would mean that I add most pegjs types to the check loop. I'll check it out and get back to you, but I think using a similar approach here is a-bit of an overkill, especially when the grammer is large.

@dmajda I don't mind splitting it in 2, I only put them together because it would be less work ;P

dmajda:
No activity here for some months ⇒ closing. If you will work on the duplicate labels check in the future, please open a new PR.

PR #334, opened 2015-04-10
Replace deprecated util.puts and util.error

burningtree:
In the newest versions of node.js (>0.12) and io.js is `util.puts` and `util.error` deprecated. Now, if you use these functions, runtime throw to stdout unnecessary message like `util.puts: Use console.log instead`.

Reference:
https://nodejs.org/api/util.html#util_util_puts
https://iojs.org/api/util.html#util_util_puts

How to reproduce:
```
$ node --version
v0.12.2

$ pegjs --version
util.puts: Use console.log instead
PEG.js 0.8.0
```

This pull request solve this problem. I updated also devDependencies because in older versions of jasmine these functions have also been used.


dmajda:
@burningtree Sorry I got to this so late. If you’re still interested, could you please:

  1. Update your `devDependencies` commit with latest versions of the packages as of now.
  2. Squash the last two commits. (Having a commit history with a commit that breaks formatting and another one that fixes it makes no sense.)
  3. Remove commas at the end of commit messages to fit with the style PEG.js uses.

With these changes I’d be happy to merge.

dmajda:
No reply, so I [replaced](https://github.com/pegjs/pegjs/commit/de1704f007523da4e8309d799adc24e9adcb0ea7) the `util.puts` and `util.error` calls myself. I’ll update `devDependencies` just before the 0.9.0 release. Closing.

PR #335, opened 2015-04-19
Generate labels for unlabelled match results

johnsoftek:
I am using pegjs to power a generalised source code formatter. The formatter needs access to all input tokens, including white space. Rather than label every match result, I would like to have access to all match results in a parser action.

One way to achieve this is to automatically supply a label for each unlabelled match result. For example, this rule
```
set_item
  = column __ KW_ASSIGN __ additive_expr {
      return { type: 'set_item', value: arguments} )
    }
```
would call the parser action with parameters _1, _2, _3, _4, _5 so the special arguments object will contain all matched results.

I may sometimes need access to individual match results:
```
comparison_expr
  = left:additive_expr __ rh:comparison_op_right? {
      if (!rh) {
        return left;
      } else {
        return arguments;
      }
    }
```
will call the parser action with parameters left, _2, right.

I do not expect this pull request to be merged. I just wished to illustrate one possible solution. I am sure there are other ways to satisfy my requirements.

(By the way, most tests pass, except the bytecode tests because of the unlabelled nod type.)

### DIFF
--- a/lib/compiler/passes/generate-bytecode.js
+++ b/lib/compiler/passes/generate-bytecode.js
@@ -436,6 +436,19 @@ function generateBytecode(ast) {
       });
     },
 
+    unlabeled: function(node, context) {
+      var env = objects.clone(context.env);
+      var label = '_' + (Object.keys(context.env).length + 1);
+
+      context.env[label] = context.sp + 1;
+
+      return generate(node.expression, {
+        sp:     context.sp,
+        env:    env,
+        action: null
+      });
+    },
+
     text: function(node, context) {
       return buildSequence(
         [op.PUSH_CURR_POS],
--- a/lib/compiler/visitor.js
+++ b/lib/compiler/visitor.js
@@ -46,6 +46,7 @@ var visitor = {
           action:       visitExpression,
           sequence:     visitChildren("elements"),
           labeled:      visitExpression,
+          unlabeled:    visitExpression,
           text:         visitExpression,
           simple_and:   visitExpression,
           simple_not:   visitExpression,
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -31,7 +31,7 @@ module.exports = (function() {
         peg$startRuleFunctions = { Grammar: peg$parseGrammar },
         peg$startRuleFunction  = peg$parseGrammar,
 
-        peg$c0 = function(initializer, rules) {
+        peg$c0 = function(_1, initializer, rules) {
               return {
                 type:        "grammar",
                 initializer: extractOptional(initializer, 0),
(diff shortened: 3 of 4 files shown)


futagoza:
Apart from the auto naming (e.g _1, _2, etc) this is actually a good idea as it would allow us to have unnamed arguments that are created from an entire rule expression and passed onto the actual action:
```js
start = int "+" int { return arguments; }
int = i:$([0-9]*) { return parseInt(i,10) }
```
The above, when passed `5+5`, should return `[5, "+", 5]`
This would help make the grammer more readable when the above behaviour is intended, as currently this is the way:
```js
start = a:int b:"+" c:int { return arguments; }
int = i:$([0-9]*) { return parseInt(i,10) }
```
Also, to make it more readable, we could take @johnsoftek idea but instead of using `_` use  `$` as they are more often used in argument handling situations like this:
```js
start = int "+" int { return arguments; }
int = $([0-9]*) { return parseInt($1,10) }
```
What do you this @dmajda?

johnsoftek:
@futagoza Yes. I often need all match results as well as access to specific match results.

Providing a parameter for each unlabelled result is not the only solution; it was just the easiest one for me to implement at the time. A better solution would be just one additional parameter, say $all, that contained an array of all match results. For example,
```
comparison_expr
  = left:additive_expr __ rh:comparison_op_right? {
      if (!rh) {
        return left;
      } else {
        return $all;
      }
    }
```
In this case, the parser would generate a function call with parameters ```(left, rh, $all)```

```$all``` would be preferable to ```arguments``` because it is a true array. Also, it aligns with current parser behaviour: with no parser action, the rule returns multiple match results as an array.

futagoza:
Your'e right, `arguments` isn't an array but as is the case, most of the time JSON AST's are the primary generated code from parsers made in PEG.js, and having the parser do a `Object (arguments)` to `Array ($all)` conversion each time an action is called will not be an optimal choice as `arguments` would be converted to an array anyway by `JSON.stringify`, therefore I think it would be better leaving the `$all` out and letting the user deal with `arguments`

johnsoftek:
@futagoza. ```arguments``` is generated by the javascript engine for the caller. The parser, as the caller, has nothing to do with the arguments variable and could not use it to generate a $all parameter.

(And, by the way, you are suggesting a conversion to array by JSON.stringify.)

If the PEG parser generated code to add a $all parameter to the parser action function call, there would be no need for the parser action to convert ```arguments```.

Performance should not dictate good design. IMO, $all is a reasonable addition. Maybe it could be turned on by a PEG parser config option.

johnsoftek:
@majda Your opinion please.

dmajda:
@johnsoftek The problem I see with having a way to access results of unlabeled expressions is that it blocks a possible optimization path. If we know that some expression is unlabeled is thus not accessed in an action or preducate, the parser doesn’t have to build a match result for it, which can lead to significant time savings. This path isn’t currently exploited but I certainly want to exploit it at some point before 1.0.0.

An obvious solution is hiding the feature behind an option but I don’t like that at all (the less options and semantic variations PEG.js has, the better).

I’ll close this PR now (as I won’t be merging it) but if you like, feel free to create an issue for the problem. Maybe there is some solution which would have a good performance and still satisfy your needs. Just be advised that I probably won’t be actively looking into this at least until 1.0.0 is released.

PR #336, opened 2015-04-30
Simplified Whitspace handling in the Pegjs Source Grammar

kwesibrunee:
The Peg Source Grammar suffers from tons of __ rules scattered throughout. This pull request changes it so that it runs the __ before all tokens. This greatly simplifies the grammars readability.

There may also be a slight performance boost as there were some unnecessary __ rules in the original grammar.

### DIFF
--- a/src/parser.pegjs
+++ b/src/parser.pegjs
@@ -70,16 +70,19 @@
   function buildList(first, rest, index) {
     return [first].concat(extractList(rest, index));
   }
+  
 }
+
+
 
 /* ---- Syntactic Grammar ----- */
 
 Grammar
-  = __ initializer:(Initializer __)? rules:(Rule __)+ {
+  =  initializer:Initializer? rules:Rule+  {
       return {
         type:        "grammar",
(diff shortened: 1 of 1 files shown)


dmajda:
@kwesibrunee Thanks for your contribution, but I won’t merge it for two reasons:

  1. I consider putting `__` before/after every token slightly hacky and putting it explicitly in the rules “only” somewhat annoying. Admittedly, this is an issue of taste.
  2. More importantly, parsing whitespace as part of tokens negatively influences location information (and thus also error reporting). For example, you modify the `AnyMatcher` rule like this:

    ```
    AnyMatcher
      = __ "." { return { type: "any", location: location() }; }
    ```

     This (together with your other changes) means in grammars like `start = /* foo */ .`, the AST node representing `.` will contain location info indicating it spans from right after the `=` character to the end of input. But this is not what is desired,  the span should cover just the `.` character. For `AnyMatcher` this doesn’t really matter (I chose that example just for simplicity), but in other places it does.

Issue #324, opened 2015-02-24
Make generated parsers use strict mode


dmajda:
This means putting the `"use strict"` directive to the toplevel of the generated parser.

dmajda:
Fixed by #339.

Issue #323, opened 2015-02-24
Convert PEG.js code to strict mode


dmajda:
This means putting the `"use strict"` directive into each JavaScript file and to the toplevel of the generated browser version.

dmajda:
Fixed by #345.

PR #339, opened 2015-06-04
"use strict";

arlolra:
* Issues #323 and #324

 * The changes to the tests probably aren't what's desired considering
   what they're trying to test, but, "in strict mode code, functions
   can only be declared at top level or immediately within another
   function.


dmajda:
@arlolra Thanks for your contribution. However, it looks like you’re mixing several things in one commit:

* Fixing issues preventing using strict mode in generated parsers and enabling it there (#323).
* Fixing issues preventing using strict mode in PEG.js code and enabling enabling it there  (#324).
* Bunch of other unrelated changes (e.g. `.jshintrc` edits).

Can you please split the PR into one solving #323, one solving #324, and then one for each related set of other changes? This will allow me to review and merge the changes more easily. Also please create a separate commit for each logical change (e.g. in #323 and #324 PRs, there should be a commit for each kind of change that prevents enabling strict mode and then one commit for actually enabling it).

I’d suggest to proceed sequentially and start with #324 because that’s probably the easiest part.

arlolra:
Sure, no problem.

PR #341, opened 2015-06-05
Split up the work from #339

arlolra:


dmajda:
It would be great to separate the “run `make hint`” change from the “test against other Node versions” change — they are logically separate.

dmajda:
Added a bunch of comments (which mostly disappeared meanwhile — you apparently push-forced). With the mentioned issues fixed I’d be glad to merge.

arlolra:
Should be as desired now.

dmajda:
Thanks!

Issue #331, opened 2015-04-08
SyntaxError has undefined `stack` property


yungsters:
It would be useful if the SyntaxError thrown by pegjs had the `stack` property:

```
this.stack = new Error().stack;
```

dmajda:
Makes sense, custom errors should not be distinguishable from native ones in this regard. Some reference material:

* [`Error.prototype.stack` documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack)
* [Inheriting from the Error object - where is the message property?](http://stackoverflow.com/questions/8802845/inheriting-from-the-error-object-where-is-the-message-property)

It would be great if you or anyone could create a patch. My requirements:

* Works for both `GrammarError` (thrown by `PEG.buildParser`) and `SyntaxError` (thrown by generated parsers).
* Works in environments where `Error.prototype.stack` is not supported (in particular, IE 8-9).
* The stack trace looks exactly the same as if `Error` was thrown instead of `SyntaxError`/`GrammarError`.

mbaumgartl:
I implemented a fix which adds the missing stack property. But I didn't pay attention to IE 8-9 so far. In my current implementation the property is set to undefined.

What is the expected behaviour if the browser doesn't support a feature?

dmajda:
> What is the expected behaviour if a browser doesn't support a feature?

@mbaumgartl I’ll quote myself:

> Makes sense, custom errors should not be distinguishable from native ones in this regard.

So if IE sets `stack` to `"bananas"` when `new Error` is called, PEG.js should do the same. If it doesn’t set it at all, PEG.js should do the same. And so on, across all supported IE versions.

PR #342, opened 2015-06-05
Add stack property to custom errors

mbaumgartl:
`Error.captureStackTrace` is a only available in [V8](https://code.google.com/p/v8-wiki/wiki/JavaScriptStackTraceApi) and will be used in Chrome and Node.js.

Other JavaScript engines will copy the stack property (if available) from a native `Error` object. The error stack is created if the property is used the first time. Accessing the stack property using a getter method/function defined by `Object.defineProperty` is faster because the error stack is only built if the property is accessed.


dmajda:
@mbaumgartl Thanks! I’ll have a look soon.

dmajda:
@mbaumgartl I extracted the core code (see [this Gist](https://gist.github.com/dmajda/fe2ed00ba59bd9ebd278)) and tested in various environments I have available:

**Node.js 0.10.38**

  *  The stack gets attached for `Error` and `StackError`, not for `NoStackError`. This means the approach works as designed here.
  * On uncaught error, the stack gets displayed for `Error` and `StackError`, not for `NoStackError`. This means Node.js probably uses the `stack` property here.

**Chrome 43.0.2357.124**

  *  The stack gets attached  for `Error` and `StackError`, not for `NoStackErrror`. This means the approach works as designed here.
  * On uncaught error, the stack is displayed in the developer tools in all cases, but differently in case of `StackError` and `NoStackError` than for `Error`. This suggests presence of some internal magic.

**Firefox 38.0.5**

  * The stack gets attached  for `Error` and `StackError`, not for `NoStackErrror`. However, the stack for `StackError` contains an additional line for the constructor call. This is a violation of the [requirements](https://github.com/pegjs/pegjs/issues/331#issuecomment-91124189) I outlined in #331.
  * On uncaught error, the stack isn’t displayed in the developer tools in any case.

**IE 11.0.9600.17420**

  * The stack gets attached  for `Error` and `NoStackError`, not for `StackErrror`. This means IE does things right by its own and we’re somehow breaking its internal mechanism.
  * On uncaught error, the stack isn’t displayed in the developer tools in any case.

Given these results, brittleness of the stack-adding code, and even bigger brittelness of any code that would try to fix its deficiencies, I think it would make most sense to simplify it just to:

```javascript
if (typeof Error.captureStackTrace === "function") {
  Error.captureStackTrace(this, GrammarError);
}
```

(Similarly for `PEG$SyntaxError`.)

That way Node.js and Chrome wiill work the same and IE will no longer be broken. The victim is Firefox, but I think I can live with that.

Can you please adapt your PR this way (or create a new one)? Also please:

  * Add “Fixes #331.” to the commit message so the issue auto-closes on merge.
  * Remove the test as it will be unreliable.

Thanks!

mbaumgartl:
@dmajda: Thanks have the quick code review!

If I simplify the code as mentioned above, I don't get any stack for StackError objects in Firefox. Can you confirm this, please?

dmajda:
> If I simplify the code as mentioned above, I don't get any stack for StackError objects in Firefox. Can you confirm this, please?

Yes, that’s the same behavior I got in my experiments — and why I wrote that Firefox is a victim.

mbaumgartl:
Tests are removed and code is simplified.

dmajda:
Can you please remove the empty line added at the top of `peg$SyntaxError` and the empty line and the `return this;` statement added at its bottom? Also, can you squash your commits into one? I’d be happy to merge then.

PR #343, opened 2015-06-05
More split outs from #339

arlolra:


arlolra:
Ok, these commits should complete #324.

I'll start on the next after review.

dmajda:
Looks good beside the few nits I mentioned.

arlolra:
Fixed the nits. Thanks for the review.

dmajda:
That was quick :-) Merging…

PR #345, opened 2015-06-08
Even more split outs from #339

arlolra:


arlolra:
Ok, that should be it for #323

dmajda:
Can you please split any changes not directly related to enabling strict mode into separate PR(s)? (Where *directly related* = without them code doesn’t work properly or JSHint produces errors.)

arlolra:
Sure, coming right up.

arlolra:
Ok, updated the branch.

dmajda:
Thanks for the work, I commented couple of minor issues.

arlolra:
Thanks for the review. I'll have these issues sorted shortly.

Can you see my [comment](https://github.com/pegjs/pegjs/pull/345#commitcomment-11658283) above about the leading `;`.

dmajda:
FYI I just realized there is an older PR (#334) implementing essentially the same changes that were part of this PR before I asked for focusing on #323. I’ll work with the author to get them in because he was first but if that gets stale, a PR from you is still welcome.

arlolra:
Should be better now.

I left in the `node` directive in jshint and explicitly disabled it for `benchmark/index.js`. Otherwise, the majority of the files complain. Happy to change that if you prefer it another way.

dmajda:
> Should be better now.

Looks great, thanks! Merging in a second.

> I left in the `node` directive in jshint and explicitly disabled it for `benchmark/index.js`. Otherwise, the majority of the files complain. Happy to change that if you prefer it another way.

Makes sense to me.

dmajda:
@arlolra As a side note, I’m on vacation starting tomorrow, so if you come up with more PRs I may react slowly or not at all for the next 2-3 weeks.

arlolra:
Thanks, and have a nice vacation! I'll probably work on the interruptible parser plugin in the meantime.

Issue #89, opened 2012-05-28
Curly braces in JavaScript code must be balanced


Dandandan:
I am writing something that compiles to JSON. Simple code:

```
start
  = a

a
  = x:(.*) {return x.join("{")}
```

doesn't work.

workaround is:

```
  = x:(.*) {return x.join(String.fromCharCode(123))}
```

(or 125 for other brace)

felix9:
I'd be inclined to close this as a wontfix, because tokenizing js requires parsing js (due to the / ambiguity and automatic semicolon insertion), and adding a js parser seems kind of heavy.

I've submitted two small PRs that should ameliorate this problem.
#534 improves the error message for unbalanced braces.
#533 documents the simple workaround.

Mingun:
> I'd be inclined to close this as a wontfix

I think, after implementing #256 this type of errors can be eliminated, by the specifying language for actions in grammar through annotations. In this case for each language at parse of a code block it is possible to attach specialized (perhaps, simplified) parser which will look for the end of the code block taking into account escaping rules for curly braces

felix9:
I'm really wary of a promise to support arbitrary language parsers. On the one hand, every supported language becomes a maintenance burden. On the other hand, there are always more languages that are unsupported, which is a development burden.

How about instead, generalize the syntax for code blocks to be delimited by n consecutive braces.

So if a code block starts with `{{` it's only closed by `}}`. '{{{{' can only be closed by '}}}}', etc.

bd82:
### Outside the box / Feature suggestion
This idea may be too much "outside the box" but please hear me out:

Perhaps this whole issue would become a-lot less important if the whole concept of **embedded
code blocks** would not be the only way (or even the default way) to add semantics to a pegjs grammar.

Some related references from similar tools:
* [Antlr4 Parse Tree Listeners](https://github.com/antlr/antlr4/blob/master/doc/listeners.md)
   > Listeners and visitors are great because they keep application-specific code out of grammars, making grammars easier to read and preventing them from getting entangled with a particular application.

* [Ohm's Philosophy](https://github.com/harc/ohm/blob/master/doc/philosophy.md)
   > Grammars are pure, not cluttered with semantic actions. 
   
   - [Math example](https://github.com/harc/ohm/blob/master/examples/math/index.html)

* [Chevrotain's CST Docs](https://github.com/SAP/chevrotain/blob/master/docs/concrete_syntax_tree.md)
[...]

PR #353, opened 2015-07-09
Allow unbalanced braces

### LINKED ISSUE #352: Rules can't contain quoted braces unless they are balanced
The grammar for peg.js rules says:
```
SourceCharacter
  = .

CodeBlock "code block"
  = "{" code:Code "}" { return code; }

Code
  = $((![{}] SourceCharacter)+ / "{" Code "}")*
```

This means that the following causes a syntax error:
```
myrule = "{" { return "{"; }
```
because the quoted brace is unbalanced.

A fix would be something like:
```
Code
  = $(Comment / StringLiteral / (![{}"'] SourceCharacter)+ / "{" Code "}")*
```
Which would allow braces and quotes (even unbalanced ones!) in comments or string literals.

arlolra:
* Fixes #352


cscott:
Look good, I like the test cases.

The RegExp grammar might be a little bit of overkill, you don't need to parse regexps that precisely, you just need to parse enough to ensure you can find the closing delimiter.  For example, there's no need to parse the flags, that can just get swallowed by the general SourceCharacter production inside Code.  You can't have any special characters or escapes in the flags, so you don't need to actually parse them.

That said, sticking to the ES6 production names (as you have) will make it easier for readers to follow the grammar, so that's probably worth a bit of spurious precision.

+1

dmajda:
Your code doesn’t handle cases like this (the action is a perfectly valid JavaScript):

    start = "a" { return ({ a: 1 / 2 }) / 3; }

A more sophisticated approach is needed.

In any case, #89 is marked as post-1.0.0, so I won’t spend time on this now, sorry.

cscott:
If you are concerned about that, just drop the RegExpLiteral production from the patch.   That would still handle all unbalanced braces that don't occur in regexps, which is much better than the current situation!

cscott:
jslint uses a simple rule (http://stackoverflow.com/a/11766233), I'll update this patch to do the same.

Issue #373, opened 2015-09-11
Consider using pure-cjs for building the browser version


dmajda:
The current way of building browser version is ugly, let’s offload that to some dedicated tool such as [pure-cjs](https://github.com/RReverser/pure-cjs).

Issue #362, opened 2015-08-08
AMD support


dmajda:
PEG.js should support AMD. While I was reluctant to adding this in the past, AMD is still used in the wild and it certainly will be used for some time (most likely until ES6 modules get widespread browser support). This means supporting AMD will benefit PEG.js users. The best way implement this is via [UMD](https://github.com/umdjs/umd).

This issue captures two separate but related tasks:

  1. Making PEG.js itself UMD module.
  2. Making generated parsers UMD modules.

Note there must be a way to specify dependencies for generated parsers. A  `dependencies` option to `PEG.buildParser` would do the trick. It would take an object mapping dependency paths to names of variables used in the code, or just an array listing the dependency names (in case the dependency paths are the same as variable names). In the CLI, the same thing could be accomplished using a `--dependencies` option taking a comma-separated list of dependencies, each of which would be either a name or a path/variable name pair (separated by `:`).

Reference: #64

## Checklist

  - [x] Make PEG.js itself UMD module
  - [x] Move module system handling from `bin/pegjs` to `buildParser`
  - [x] Make generated parsers UMD modules
  - [x] Implement a way to specify parser dependencies in a module system independent fashion

seeden:
Maybe the better idea is to use ES6 syntax for your project. And then automatically create commonjs/amd/umd library. For this purpose I can recommend you Babel and Webpack. Babel will create automatically ES5 version from your ES6. And webpack will create bundled library.

dmajda:
ES6 & Babel: I’ll probably have a look at some point, but right now I don’t see that many benefits for PEG.js to warrant complicating the infrastructure and losing control over the exact shape of the code.

webpack: AFAICT, this is something more useful for websites rather than libraries. Also, I’ll touch it only if there is no other choice (it looks like a tangle of bad design from a superficial look).

dmajda:
This is implemented now. Compared to the ideas outlined in the description, there were two changes:

  * The `dependencies` option of `PEG.buildParser` can’t take an array, just an object. This makes things a bit simpler.

  * The `--dependencies` CLI option which was supposed to describe the whole dependency list was replaced by a `-d`/`--dependency` option which describes an individual dependency (it can be specified multiple times). The advantages of this approach are:

    * It avoids handling of comma-separated lists (which could get messy quickly with more complex module IDs).
    * It is similar to already existing `--plugin` option.
    * It is similar to specifying dependencies in other environments (e.g. `--require` in Node.js).

PR #379, opened 2015-10-13
Use browserify for building the browser version

joseph-onsip:
This resolves https://github.com/pegjs/pegjs/issues/373 and,
since `browserify` produces a UMD bundle (due to `--standalone PEG`),
addresses the first part of https://github.com/pegjs/pegjs/issues/362,

> 1. Making PEG.js itself UMD module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,27 +2,6 @@
 
 PEGJS_VERSION = `cat $(VERSION_FILE)`
 
-# ===== Modules =====
-
-# Order matters -- dependencies must be listed before modules dependent on them.
-MODULES = utils/arrays                          \
-          utils/objects                         \
-          utils/classes                         \
-          grammar-error                         \
-          parser                                \
-          compiler/visitor                      \
-          compiler/asts                         \
-          compiler/opcodes                      \
-          compiler/js                           \
-          compiler/passes/generate-bytecode     \
-          compiler/passes/generate-js           \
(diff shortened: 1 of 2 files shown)


dmajda:
@joseph-onsip Nice patch, thanks! Could you please fix the small nit I pointed out in the [other comment](https://github.com/pegjs/pegjs/pull/379#commitcomment-13817973) and rebase (so the change is still one commit)? I’ll merge then.

I also compared browser build sizes with the current code and after applying your patch:

**Before**

```
282660 peg-0.9.0.js
122389 peg-0.9.0.min.js
``` 

**After**

```
256077 peg-0.9.0.js
123484 peg-0.9.0.min.js
```

The difference in case of `peg-0.9.0.js` is caused mostly by indentation whitespace (the current code indents the modules, Browserify does not). The difference in case of `peg-0.9.0.min.js` is Browserify’s overhead. Although I’m not much happy about it I think it is small enough to live with it.

joseph-onsip:
Sure thing, I've added `MAIN_FILE` and squashed all changes into 0d8c0458232339e2bc399c01140561d14e6a8058.

I also noticed the change in build sizes, and was a little irked by the increase in `peg-0.9.0.min.js`, but after realizing that it's also producing a UMD module (for #362), I figured it was worth it.

dmajda:
Thanks, merging now.

PR #380, opened 2015-10-16
travis: use containers for faster builds

joseph-onsip:
http://docs.travis-ci.com/user/migrating-from-legacy/


dmajda:
Nice, thanks!

PR #381, opened 2015-10-18
Make generated parsers UMD modules using `umd` 3.0.1

josephfrazier:
[`umd`][1] is what `browserify` uses (via [`browser-pack`][2]),
to generate UMD modules, so it seems appropriate to use here as well.

addresses #362

[1]: https://www.npmjs.com/package/umd
[2]: https://www.npmjs.com/package/browser-pack

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,10 +68,13 @@ file but with “.js” extension. You can also specify the output file explicit
 
 If you omit both input and output file, standard input and output are used.
 
-By default, the parser object is assigned to `module.exports`, which makes the
-output a Node.js module. You can assign it to another variable by passing a
-variable name using the `-e`/`--export-var` option. This may be helpful if you
-want to use the parser in browser environment.
+The output is a [UMD](https://github.com/ForbesLindesay/umd) module that
+exports the parser. If no module system is present, the parser object is
+assigned to `peg$parser` by default. You can assign it to another variable by
+passing a variable
+[name](https://github.com/ForbesLindesay/umd/tree/3.0.1#name-casing-and-characters)
+using the `-e`/`--export-var` option. This may be helpful if you want to use
+the parser in browser environment.
 
 You can tweak the generated parser with several options:
 
--- a/bin/pegjs
+++ b/bin/pegjs
@@ -4,7 +4,10 @@
 
 var fs   = require("fs");
 var path = require("path");
+var umd = require("umd");
 var PEG  = require("../lib/peg");
+
+var DEFAULT_EXPORT_VAR = "peg$parser";
 
 /* Helpers */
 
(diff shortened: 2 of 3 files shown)


dmajda:
@josephfrazier I looked at the output of `umd` and it looks like it generates boilerplate code which is minimized and also more elaborate than [UMD’s `returnExports.js` format](https://github.com/umdjs/umd/blob/master/templates/returnExports.js) (which I think would be the most suitable here). I really want generated parser code to be readable and not contain anything unnecessary, which means `umd`  is not up to the job. Some other library or manual code generation is needed.

I’m also thinking about the user interface. I’d say it makes sense to rename `--export-var` to `--name` because it will really be a module name now. The default should be an empty string which would mean that the global variable assignment would not be not generated, only `define(...)` and `module.exports = ...`. This will force people to name their parser meaningfully if they want to use it in browser environment.

What do you think? Would you be willing to implement these changes? (I realize I’m replying quite late so if you lost interest meanwhile I perfectly understand.)

josephfrazier:
>  I’d say it makes sense to rename --export-var to --name because it will really be a module name now

Agreed.

> UMD’s returnExports.js format (which I think would be the most suitable here)

How about building/outputting [the dependency-free form of returnExports.js](https://github.com/umdjs/umd/blob/d31bb6ee7098715e019f52bdfe27b3e4bfd2b97e/templates/returnExports.js#L41-L60), but changing the `define()` call to [include the module name](http://requirejs.org/docs/whyamd.html#namedmodules) and changing the `root.returnExports` line to set `root.$MODULE_NAME` instead? As you mention, setting the global variable would only happen if the module name was provided via `--name`, whereas the AMD definition could simply be anonymous.

Another thing that comes to mind is whether the `--name` parameter should influence the output filename, since [CommonJS module names are implied by the filename](https://github.com/umdjs/umd/blob/d31bb6ee7098715e019f52bdfe27b3e4bfd2b97e/templates/returnExports.js#L7). I'm not sure if I have a good answer for this at the moment, but perhaps you have a particular direction in mind for the API?

dmajda:
Sorry for not replying here for so long.

I ultimately [implemented the UMD support myself](https://github.com/pegjs/pegjs/compare/5e702b5901abe559b5d52713c3794a77e83c331a...810567d865930f2ef379aaaffdc6d0e87ca1555c). I kept the `--export-var` option and didn’t change it to `--name` as I originally intended. The reason is that in the AMD case the parser is always defined as an anonymous module by `define` (in accordance with [best practices](http://requirejs.org/docs/whyamd.html#namedmodules)), so naming the option `--name` could lead to confusion (“I specified a module name, why wasn’t it used?”).

> Another thing that comes to mind is whether the `--name` parameter should influence the output filename, since [CommonJS module names are implied by the filename](https://github.com/umdjs/umd/blob/d31bb6ee7098715e019f52bdfe27b3e4bfd2b97e/templates/returnExports.js#L7). I'm not sure if I have a good answer for this at the moment, but perhaps you have a particular direction in mind for the API?

Given that `--export-var` doesn’t influence module loading in any way and I’m satisfied with the current output name handling (output file name derived from the input file name) I kept things as they were.

Thanks for your contribution!

PR #382, opened 2015-10-18
travis: cache node_modules for faster builds

josephfrazier:
This reduces build times by around 10 seconds. ([example](https://travis-ci.org/joseph-onsip/pegjs/builds/86058074))

http://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/


dmajda:
Thanks!

Issue #370, opened 2015-09-11
Small CLI improvements


dmajda:
* [x] Allow using `-` to mean standard input and output.
  * [x] More compiler-like interface (e.g. `-o` for output file)? Support for multiple compiled files? Look at CLI of various compiler-like tools (e.g. CoffeeScript, SASS).

See also #429.

PR #383, opened 2015-10-19
WIP: Use minimist to process CLI arguments

josephfrazier:
This is very much a work in progress, but I wanted to get some feedback. I recommend viewing the diff with github's [whitespace](https://github.com/blog/967-github-secrets#whitespace) option: https://github.com/pegjs/pegjs/pull/383/files?w=true

Using [minimist](https://www.npmjs.com/package/minimist) to process CLI arguments makes the parsing code generally more flexible, and addresses the second part of #370:

> - [ ] Allow e.g. `--export-var=foo` (in addition to `--export-var foo`).

In addition, ba54ad03f5afb3710fe19e724a578e9c3a0e768e addresses the first part of #370:

> - [ ] Allow using `-` to mean standard input and output.

So far, my concern is that I might have accidentally broken something, given that there currently aren't any automated tests for the CLI.

### DIFF
--- a/bin/pegjs
+++ b/bin/pegjs
@@ -4,6 +4,8 @@
 
 var fs   = require("fs");
 var path = require("path");
+var minimist = require("minimist");
+var arrays  = require("../lib/utils/arrays");
 var PEG  = require("../lib/peg");
 
 /* Helpers */
@@ -88,15 +90,10 @@ function trim(s) {
 
 /* Arguments */
 
-var args = process.argv.slice(2); // Trim "node" and the script path.
-
-function isOption(arg) {
-  return (/^-/).test(arg);
-}
-
-function nextArg() {
-  args.shift();
-}
+// Trim "node" and the script path.
+var argv = minimist(process.argv.slice(2), {
+  "boolean": ["cache", "trace"]
+});
 
 /* Files */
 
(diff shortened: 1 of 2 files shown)


dmajda:
@josephfrazier Thanks a lot for the effort, however I don’t think using minimist is a good idea. Its API is ugly and little too much minimal, which results in code I find worse than the original version.

I did a research on option parsing libraries for Node.js a while ago and the only library which passed my bar is [yargs](https://www.npmjs.com/package/yargs). Its default chaining API is horrible, but using `.options` allows a declarative way of specifying option which I think would work fairly well. It also looks pretty mature and under active development so there is fairly low risk of painting PEG.js into a corner.

Let me know if you’d like to tackle rewriting `bin/pegjs` using yargs instead of minimist. If not, I’ll do that at some point.

(As a side note, my ideal option parsing API would look something like Ruby’s [`OptionParser`](http://ruby-doc.org/stdlib-2.2.3/libdoc/optparse/rdoc/OptionParser.html). But I didn’t find a npm in this style with the maturity of yargs.)

josephfrazier:
No worries, @dmajda; I was also using this to acquaint myself with minimist. I might be interested in doing the yargs rewrite, but I'd first like to write some tests for the command line API. Would be you open to considering a pull request with tests that use [child_process](https://nodejs.org/docs/v0.12.9/api/child_process.html) to ensure that `bin/pegjs` behaves as expected?

PR #386, opened 2015-11-05
Create a byte code optimiser

pwagland:
Add the first stab at creating a byte code optimiser. This will optimise out if statements that it can determine will always be true. This typically happens after an option block, since the the stack pointer is set to 0 if it is in error, and the next statement typically checks for error. That check, and the else branch, is now elided.


pwagland:
Please see https://github.com/pegjs/pegjs/pull/387 instead. That comes from a different branch than master.

PR #387, opened 2015-11-06
Add byte code optimizer to eliminate redundant checks

pwagland:
Add the first stab at creating a byte code optimiser. This will optimise out if statements that it can determine will always be true. This typically happens after an option block, since the the stack pointer is set to 0 if it is in error, and the next statement typically checks for error. That check, and the else branch, is now elided.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -22,6 +22,7 @@ var compiler = {
     },
     generate: {
       generateBytecode:    require("./compiler/passes/generate-bytecode"),
+      optimiseBytecode:    require("./compiler/passes/optimise-bytecode"),
       generateJS:          require("./compiler/passes/generate-js")
     }
   },
--- a/lib/compiler/passes/optimise-bytecode.js
+++ b/lib/compiler/passes/optimise-bytecode.js
@@ -1,0 +1,370 @@
+"use strict";
+
+var arrays = require("../../utils/arrays"),
+    asts   = require("../asts"),
+    op     = require("../opcodes"),
+    js     = require("../js");
+
+/* Generates parser JavaScript code. */
+function optimiseBytecode(ast, options) {
+
+  var varType = {
+    UNKNOWN    : "Unknown",
+    FAILED     : "Failed",
+    NOT_FAILED : "Not_failed"
+  }
+
+  function optimiseRuleFunction(rule) {
+    var stack = {
+      types : [],
+      push: function(exprCode) {
+        if (/^null$/.test(exprCode)) {
+          this.types.push(varType.NOT_FAILED);
+        } else if (/^peg\$FAILED$/.test(exprCode)) {
+          this.types.push(varType.FAILED);
+        } else {
+          this.types.push(varType.UNKNOWN);
+        }
+      },
+      pop:  function() {
+        if (arguments.length === 0) {
+          return this.types.pop();
+        } else {
+          var n = arguments[0], types = [];
+          while (n-- > 0)
+          {
+            types.push(this.types.pop());
+          }
+          return types;
+        }
+      },
+      top: function() {
+        return this.types.length - 1;
+      },
+      index: function(i) {
+        return this.types.length - 1 - i;
+      }
+    }
+
+    function compile(bc) {
+      var ip = 0,
+          end = bc.length,
+          newbc = [],
+          value;
+
+      function pushToNewBC(item) {
+        return newbc.push(item);
+      }
+
+      // This is the condition compilation function when we can work out
+      // what the value of the condition will be.
+      function compileConditionFail(cond, fail, argCount) {
+        var baseLength = argCount + 3,
+            thenLength = bc[ip + baseLength - 2],
+            elseLength = bc[ip + baseLength - 1],
+            baseIp     = ip,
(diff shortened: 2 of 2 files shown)


dmajda:
@pwagland I finally had time to look at this.

You don’t say how much faster generated parsers become with this change so I ran it through [`tools/impact`](https://github.com/pegjs/pegjs/blob/master/tools/impact):

```
Speed impact
------------
Before:     736.37 kB/s
After:      821.90 kB/s
Difference: 11.61%

Size impact
-----------
Before:     721156 b
After:      687793 b
Difference: -4.63%

(Measured by /tools/impact with Node.js v4.2.4 on Darwin 15.2.0 x86_64.)
```

A speedup of 11–12% (the actual value fluctuated a bit on multiple runs) certainly makes the optimization worthwhile.

---

Now to the implementation. PEG.js bytecode wasn’t really designed to be manipulated, it was mostly meant as a low-level format which would make code generator(s) more flexible and easier to write. As a result, your code pretty much necessarily duplicates a lot of logic from the generator (including fragile things like length and offset computations) and contains some hackish parts (like `exprCode` parsing).

This made me think whether bytecode is the right level to implement this optimization. It would be probably easier to do it already when the bytecode is generated. I imagine doing it these two steps:

  1. A dedicated pass would tag expressions in the AST with a flags describing whether they always/never match (the flag could be names, say, `alwaysMatches` and `neverMatches`).
  2. The bytecode generator would then skip generating branches which are not needed based on the flags.

This solution would avoid fragile bytecode tweaking and would be conceptually simpler (I think).

Now, one can argue that this is an optimization and adding it to the generator would complicate it, violate the single responsibility principle, etc. I actually tend to agree, but I still consider this approach better than the bytecode one. It think the bytecode-based approach would make sense only if we do more optimizations on that level over time, in which case it I’d say let’s invest into uplifting the bytecode representation into something more amenable to manipulation (probably some kind of lower-level AST) and then implement the optimization using that. So my question for you is — do you have any other bytecode-level optimizations in mind, or was this a one-off?

In any case, thanks for the work so far!

pwagland:
@dmajda As to the implementation. Yeah, Your point about it being more or less a copy is a valid one. I had initially aimed to write the optimisations in the original parser, and went a fair way down that road. In the end though, the order of evaluation killed me, since at the point that we are generating byte code, we don't know if the results will be used or not, nor do we know what the previous results are. This was happening since the code is, more or less, generated inside out. This is the reason that I created a new pass, since then I could just walk over the tree in the more or less expected fashion.

That said, this could possibly be solved at generation time by passing more information around, but that didn't jump out at me as to how to do it.

I fully agree that the current code feels fragile. It was quite fragile to write as well, and without a fairly comprehensive test set I would not be sure that I had gotten it right. That said, all of the code generators need to deal with the same level of fragility at this point in time. So any pain suffered in this optimiser will also be suffered by the various 3rd party generators as well. This is also one of the reasons that I didn't implement the next optimisation, as that would have required a new op code, with the ensuing knock on effects.

As to your second question, are there anymore byte code optimisations that I wanted to… yes there was one more, but that will have less impact on the generated code, and would have require a new byte code, which would break all of the current downstream generators. The optimisation was to simply not store results that we know are not going to be further used. There are a number of times that the result is stored, in case it will be used in a function, but it then isn't. Not storing it would save generated code size, and possibly (but unlikely IMO) a little speed. IMO it would make the generated code a little simpler to read, since you would have fewer variables that you need to track to understand the behaviour of it.

The new byte code is needed for this to be able to call a method, but not store the result. At the moment, that is not possible, since the call method byte code implicitly also stores the result.

These two improvements would also knock off some, but not all AFAICT, of the improvements that were discussed in [the Google group](https://groups.google.com/forum/#!topic/pegjs/fL57bEdeuo4).

dmajda:
Thanks for your detailed response.

> That said, all of the code generators need to deal with the same level of fragility at this point in time. So any pain suffered in this optimiser will also be suffered by the various 3rd party generators as well.

Good point. The uplifting I talked about probably makes sense even without more optimizations on the bytecode level.

> The optimisation was to simply not store results that we know are not going to be further used.

Yes, this is an obvious one which I was also thinking about doing (but I postponed that as I intend to [focus on optimizations later](https://github.com/pegjs/pegjs/wiki/Roadmap#0110-performance)). Again, this can be done on the AST level since the AST nodes know whether they will use results returned by their child nodes or not. (Actually, this is a bit more complicated, since rules can be used in different contexts where sometimes the result is needed and sometimes not. So two versions of bytecode may need to be generated in some cases.)

---

Since all the optimizations discussed here apparently can be implemented on the AST level, which I much prefer, I won’t merge the bytecode level implementation and will close this PR. As I said, I consider the optimization worthwhile. If you’d like to reimplement it the AST way, I’m definitely interested and will have a look.

pwagland:
> As I said, I consider the optimization worthwhile. If you’d like to reimplement it the AST way, I’m definitely interested and will have a look.

Having already looked at doing it through the AST, I'll warn up front that it feels to me like it is going to be a bigger, albeit cleaner, change. Depending on time et al. I might have a look, I'll drop you a message if/when I decide to look into it.

PR #393, opened 2015-12-27
Make `visitExpression` default visitor function always returns value from visitor

Mingun:
This is more awaited behaviour than returning nothing.

### DIFF
--- a/lib/compiler/visitor.js
+++ b/lib/compiler/visitor.js
@@ -15,7 +15,7 @@ var visitor = {
     function visitExpression(node) {
       var extraArgs = Array.prototype.slice.call(arguments, 1);
 
-      visit.apply(null, [node.expression].concat(extraArgs));
+      return visit.apply(null, [node.expression].concat(extraArgs));
     }
 
     function visitChildren(property) {


dmajda:
> This is more awaited behaviour than returning nothing.

I don’t think so.

The overall idea is that there is no good default return value for leaf nodes (`literal`, `class`, etc.) and for complex nodes (`grammar`, `choice`, etc.). Therefore their visit functions don’t return anything (or, more exactly, return `undefined`). Visit functions for nodes containing just one expression are simply consistent with these.

Now, one could argue that visitor functions for nodes like `choice` and `sequence` could return visit results of their children as an array and that `grammar` could similarly return an object with `initializer` and `rules` properties, but I think this is a bad idea for two reasons:

  1. The structure of returned results would be inconsistent in cases a node has just one property with node(s) (e.g. `optional`, `choice`) vs. when it has more such properties (e.g. `grammar`). In the former case, the result structure would match the property, in the latter case it would match the whole node.
  2. There is performance penalty in building the results when they are not needed (e.g. the visitor does just some checks on certain nodes and doesn’t need to accumulate any result).

The structure of returned results could be made consistent (alway match the whole node), solving 1, but but this would not be practical.

Overall, I think the status quo is fine and no change is needed here.

Mingun:
> The overall idea is that there is no good default return value for leaf nodes

But for `expression`-nodes there is a good return value and I don't see the reasons why not to return it by default. If it isn't necessary for you, just don't use it. I don't think that return of excess value will a little affect perfomance, the bottleneck will be exact not here. Besides, the compiler will easily cope with optimization of this situation.

> Now, one could argue that visitor functions for nodes like  choice  and  sequence  could return visit results of their children as an array and that  grammar  could similarly return an object with  initializer  and  rules  properties, but I think this is a bad idea for two reasons:

Yes, here I agree with you and didn't change thyir visitor functions for the same reasons. 
But complex nodes considerably differ from simple nodes and shall process on a miscellaneous. That something is inconvenient for complex nodes doesn't mean that it is senseless for simple nodes.

For complex nodes if it is necessary to return something, most likely the behavior by default will already not arrange and you already have custom visitor function. For `expression` nodes, however, quite often it is required only pass arguments to and get result from child `expression` node, doing nothing in node visitor function itself, but due to the lack of a returned value it is necessary to make this logic obviously. This is boilerplate code, that can be eliminated.

Mingun:
> The overall idea is that there is no good default return value for leaf nodes

[Now I noted](https://github.com/pegjs/pegjs/blob/master/lib/compiler/visitor.js#L10):
```js
function visit(node) {
  return functions[node.type].apply(null, arguments);
}
```
For leaf/complex nodes there no good default return value, so, on your logic `return` it isn't necessary here (though it is clear that it not so. Deleting `return` will break some passes). From this point of view I don't see a difference between this function and `visitExpression`.

dmajda:
> But for `expression`-nodes there is a good return value and I don't see the reasons why not to return it by default.

The reason is consistency in returned value with other nodes, which is impossible to achieve in a practical manner (as I explained in a great detail).

> For `expression` nodes, however, quite often it is required only pass arguments to and get result from child `expression` node, doing nothing in node visitor function itself, but due to the lack of a returned value it is necessary to make this logic obviously. This is boilerplate code, that can be eliminated.

I understand this concern, but in my mind it doesn’t outweigh the need for consistency. Also, I value explicitness over relying on non-obvious implicit rules. In this case, explicitness means that in the visitor definition it is obvious that the case is handled and the intention is just to pass returned value through.

If you bump into this problem repeatedly (i.e. end up defining multiple visitors with the same boilerplate), you can just create your own set of default functions and use them as a base when calling `visitor.build`, like this:

```javascript
visitor.build(_.extend(_.clone(MY_DEFAULT_FUNCTIONS), {
   /* ... */
}));
```

> [Now I noted](https://github.com/pegjs/pegjs/blob/master/lib/compiler/visitor.js#L10):
> ```js
> function visit(node) {
>   return functions[node.type].apply(null, arguments);
> }
> ```
> For leaf/complex nodes there no good default return value, so, on your logic `return` it isn't necessary here (though it is clear that it not so. Deleting `return` will break some passes). From this point of view I don't see a difference between this function and `visitExpression`.

Well, this is the dispatcher which calls all visitor functions — including those that return values. These values need to be passed back to callers, thus the `return` statement.

The fact that default visitor functions do not return any values doesn’t mean that the visitor builder doesn’t support visitors whose functions do.

PR #397, opened 2016-01-09
Issue #9: Add testes with |trace| option set to |true|.

Mingun:
You forgot to add them. When I wrote one of the my passes, the generator broke in case of trace option on. Now such errors will be caught by tests.

### DIFF
--- a/spec/behavior/generated-parser-behavior.spec.js
+++ b/spec/behavior/generated-parser-behavior.spec.js
@@ -18,10 +18,14 @@ describe("generated parser behavior", function() {
     }
 
     var optionsVariants = [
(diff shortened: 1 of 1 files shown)


dmajda:
I agree that the `options.trace == true` case was missing, thanks for pointing it out.

Applying this PR as it is (after resolving the conflict) causes clutter in test output because the default tracer uses `console.log`. I decided to stub out `console.log` using Jasmine’s `spyOn` and I [committed the improved solution](https://github.com/pegjs/pegjs/commit/2647f5789f50ec066f6381941f3ed58606c17c0c) with a pointer to this PR.

PR #400, opened 2016-01-23
Optimize redundant fail checks

Mingun:
It is cast by the #387 -- optimization, rewrited in AST level. This PR drop unnecessary checks of parse results, if they known at compile-time. Below performance comparison. I am not sure that the script considers correctly as results strongly float. The first three runs generally showed decreased perfomance, but subsequent 7 showed improving within 3-8% (`master` is 48f5ea4b37157040edba01e098a81c7bb78cbebe). I think that it can be because of the additional pass applied when compile parser:
### First run

``` bash
$ ./impact master optimize-fail-check
Measuring commit master... OK
Measuring commit optimize-fail-check... OK

Speed impact
------------
Before:     220.35 kB/s
After:      195.96 kB/s
Difference: -11.07%

Size impact
-----------
Before:     721705 b
After:      599139 b
Difference: -16.99%

(Measured by /tools/impact with Node.js v0.10.18 on MINGW32_NT-6.0 1.0.18(0.48/3/2) i686.)
```
### Last (10) run:

``` bash
$ ./impact master optimize-fail-check
Measuring commit master... OK
Measuring commit optimize-fail-check... OK

Speed impact
------------
Before:     176.21 kB/s
After:      190.81 kB/s
Difference: 8.28%

Size impact
-----------
Before:     721705 b
After:      599139 b
Difference: -16.99%

(Measured by /tools/impact with Node.js v0.10.18 on MINGW32_NT-6.0 1.0.18(0.48/3/2) i686.)
```

### DIFF
--- a/lib/compiler/index.js
+++ b/lib/compiler/index.js
@@ -9,6 +9,7 @@ const reportDuplicateRules = require( "./passes/report-duplicate-rules" );
 const reportInfiniteRecursion = require( "./passes/report-infinite-recursion" );
 const reportInfiniteRepetition = require( "./passes/report-infinite-repetition" );
 const reportUndefinedRules = require( "./passes/report-undefined-rules" );
+const inferenceMatchResult = require( "./passes/inference-match-result" );
 const visitor = require( "./visitor" );
 
 function processOptions( options, defaults ) {
@@ -58,6 +59,7 @@ const compiler = {
         },
         generate: {
             calcReportFailures: calcReportFailures,
+            inferenceMatchResult: inferenceMatchResult,
             generateBytecode: generateBytecode,
             generateJS: generateJS
         }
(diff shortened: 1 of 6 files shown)


Mingun:
Any comments?

Mingun:
@futagoza?

futagoza:
> Actually there shall be still tests for code generation when the match property of each node is equal 1 and -1, but it requires big changes in the testing mechanism since an opportunity to set to all AST nodes the necessary property is necessary. I suppose, this general operation can be carried out as a utility method in PEG.compiler.visitor and it requires separate PR.

@Mingun I'm currently rewriting the AST visitor as well as adding some general utility methods used repeatedly throughout the library (`peg.compiler.visitor.build` will remain to help keep API compatibility for plugins and internally help migration to the new `peg.compiler.visitor.ASTVisitor` class). When I'm done and merge the changes you can check it out and see how to enhance it to help with the above

PR #403, opened 2016-01-24
Issue #402: Add new option |newLocation| when parse input to hook loc…

### LINKED ISSUE #402: Q:location() function is fine but am I able to associate this with a given input file name?
Is there a possible chance of passing the input file name to the parser so that one can associate a given parse location with that file name?

Mingun:
…ation information, collected by |location()| API function.

For example, to add filename information to all locations use this code:
```js
var parser = ...;
var filename = ...;
var grammar = fs.readFileSync(filename, 'utf-8');
parser.parse(grammar, { newLocation: function(loc) { loc.fileName = fileName; return loc; } });
```

Closes #402.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,11 @@ object to the `parse` method. The following options are supported:
 
   * `startRule` — name of the rule to start parsing from
   * `tracer` — tracer to use
+  * `newLocation` — hook function for tweak location information, attached to all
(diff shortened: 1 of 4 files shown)


dmajda:
A location-processing hook looks like unnecessarily abstract solution to me. Unless you have other important use cases in mind beyond adding a file name, I think just adding a `filename` option and adding its contents to the location into would work better. See #421 for details.

Issue #368, opened 2015-08-29
Document interaction of human-readable rule names and error handling


sampsyo:
The presence of a human-readable name for a rule (regardless of its content) can affect the kinds of errors that PEG.js produces. This is a subtle effect that is important for getting good parse errors, but it doesn't seem to be documented.

Here's a tiny grammar that parses sequences of lower-case letters separated by semicolons:

```
S = A (";" A)*
A = [a-z]
```

What happens if I mistakenly use an upper-case letter? If I try to parse `a;B`, this grammar emits the error:

```
Expected [a-z] but "B" found.
```

But if I assign a human-readable name for the first rule:

```
S "foo" = A (";" A)*
A = [a-z]
```

I get a very different error message:

```
Expected end of input but ";" found.
```

The parser seems to prefer "blaming" named rules over unnamed rules. This seems like a reasonable way to control error messages, and there's kind of an intuitive logic to it: unnamed rules are more likely to be some kind of internal implementation detail, and the library tries to avoid exposing them too much to the user.

But unless I'm missing something, this behavior isn't documented. (I found it by trial and error.) I'd be happy to draft some text to help document this if I can have some help with understanding exactly how this feature works.

dmajda:
@sampsyo You are right that this behavior is not documented, or rather documented quite vaguely:

> A rule can also contain *human-readable name* that is used in error messages (in our example, only the `integer` rule has a human-readable name).

In fact, the whole error handling area is currently not documented much.

> I'd be happy to draft some text to help document this if I can have some help with understanding exactly how this feature works.

Are you still interested in this? You pretty much already discovered how the feature works — low-level expressions (string literals, character classes, and `.`) produce *expectations* on match failures and a rule with human-readable name just overwrites these expectations with a one containing rule’s human-readable name.

sampsyo:
Sure! I won't pretend that I understand it completely, but I can try putting together a few examples to demonstrate the effect. I'll open a PR.

futagoza:
Merged #415

PR #391, opened 2015-12-12
Custom location parameter for error() and expected() fuctions

YemSalat:
#390

PR #415, opened 2016-01-31
Docs: try to explain error messages

sampsyo:
I took a shot at trying to explain the subtle error message behavior I described in #368. I'll admit that I don't really understand the design here, though, so this explanation:

> PEG.js prefers shorter matches for named rules and longer matches for unnamed rules.

is a completely made-up placeholder; it's just a guess. I'd be happy to iterate if there's more insight to incorporate here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,6 +438,47 @@ Try to match the first expression, if it does not succeed, try the second one,
 etc. Return the match result of the first successfully matched expression. If no
 expression matches, consider the match failed.
 
+Error Messages
+--------------
+
+As described above, you can annotate your grammar rules with human-readable
+names that will be used in error messages. For example, this production:
+
+    integer "integer"
+      = digits:[0-9]+
+
+will produce an error message like:
+
+> Expected integer but "a" found.
+
+when parsing a non-number, referencing the human-readable name "integer."
+Without the human-readable name, PEG.js instead uses a description of the
+character class that failed to match:
+
+> Expected [0-9] but "a" found.
+
+Aside from the text content of messages, human-readable names also have a
+subtler effect on *where* errors are reported. PEG.js prefers shorter matches
+for named rules and longer matches for unnamed rules. For example, for this
+rule matching a comma-separated list of integers:
+
+    seq
+      = integer ("," integer)*
+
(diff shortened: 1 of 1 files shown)


Mingun:
I would replace with approximately such description, with necessary corrections since it is the semi-automatic translation from Russian:
> If the rule has a human-readable name, in case of parse error when parsing this rule the describe, that all rule match failed, not just part of the rule, that actually not matched. Thus, named rules, from the point of view of error handling, are considered as the indivisible entities which don't have an inner pattern.

Original Russian description:
> Если правило имеет человеко-читаемое имя, то при возникновении ошибки разбора внутри правила ошибка ссылается на ошибку сопоставления всего правила, а не той его части, что вызвала ошибку. Таким образом, именованные правила с точки зрения обработки ошибок, рассматриваются как неделимые сущности, не имеющие внутренней структуры

sampsyo:
Thanks! That definitely helps. I'll try incorporating this now.

YemSalat:
This has been resolved here: https://github.com/pegjs/pegjs/pull/391
Now one can return custom error locations, so this pr probably does not apply anymore?

sampsyo:
The default behavior should still be documented, in my opinion.

futagoza:
@sampsyo If you can update and polish it (spelling, grammer, etc), reflecting any changes since 08/2016, that would be helpfull, and I'll merge it by the next realease then.

sampsyo:
Sure! I can take another pass, but things seem mostly in order to me. I *think* there haven't been any behavioral changes since I wrote this, but maybe someone who's been involved in development would be able to say definitively?

futagoza:
I'm going to confirm the behavioral changes since 08/2016 at a later date, but I know that the `README.md` has been updated 😄

sampsyo:
Ah, of course. I checked out the merged document and I think everything's in order. Let me know if you see anything awry.

futagoza:
Thanks 😄 

I guess I still got alot to learn about Git, just found out an hour ago how to check out a PR on a local copy of the repo 🤣 sorry about that @sampsyo

sampsyo:
No problem! Thanks for merging.

PR #439, opened 2016-08-23
Added 'kind' to VariableDeclaration in example

atavakoli:
The JavaScript example grammar's VariableDeclaration was missing the 'kind'
member (always set to "var"). Resulting AST tested with [escodegen][].

### Test code

```js
var i = 0;
```

### AST

```json
{
   "type": "Program",
   "body": [
      {
         "type": "VariableDeclaration",
         "declarations": [
            {
               "type": "VariableDeclarator",
               "id": {
                  "type": "Identifier",
                  "name": "i"
               },
               "init": {
                  "type": "Literal",
                  "value": 0
               }
            }
         ],
         "kind": "var"
      }
   ]
}
```

See [estree's spec.md][estree] for the version of the spec used. 

[estree]: https://github.com/estree/estree/blob/bc34a7cc861640aa8ee18d93186720be1b13db53/spec.md#variabledeclaration
[escodegen]: https://github.com/estools/escodegen

### DIFF
--- a/examples/javascript.pegjs
+++ b/examples/javascript.pegjs
@@ -1016,7 +1016,8 @@ VariableStatement
   = VarToken __ declarations:VariableDeclarationList EOS {
       return {
         type:         "VariableDeclaration",
-        declarations: declarations
+        declarations: declarations,
+        kind:         "var"
       };
     }
 
@@ -1124,7 +1125,8 @@ IterationStatement
         type:   "ForStatement",
         init:   {
           type:         "VariableDeclaration",
-          declarations: declarations
+          declarations: declarations,
+          kind:         "var"
         },
         test:   extractOptional(test, 0),
         update: extractOptional(update, 0),
@@ -1158,7 +1160,8 @@ IterationStatement
         type:  "ForInStatement",
         left:  {
           type:         "VariableDeclaration",
-          declarations: declarations
+          declarations: declarations,
+          kind:         "var"
         },
         right: right,
         body:  body


dmajda:
@atavakoli You are right, good catch!

I have a couple of notes to your patch:

  * There are 2 more places in the grammar where `VariableDeclaration` nodes are constructed, the `kind` property needs to be added there too.

  * The value of the `kind` property should be aligned with other values (to conform to the style used in the whole grammar).

  * The commit message should conform to [these conventions](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) (see `CONTRIBUTING.md`). Right now the subject uses incorrect tense and the body is not wrapped at 72 chars.

     It would be even better if the message looked similar to previous messages used when modifying the examples, e.g. “JavaScript example: Add "kind" to VariableDeclaration nodes” as a subject with the body describing that the property was missing. This would make the history slightly cleaner and it would be obvious from the subject which example the change affetcs.

Could you please update the patch accordingly?

atavakoli:
@dmajda Absolutely, thanks!

atavakoli:
@dmajda Does the latest commit look ok?

dmajda:
@atavakoli Perfect, thanks! Merging in a second.

I also applied [my own rule](http://majda.cz/blog/two-questions-to-ask-after-a-bugfix) and reviewed the grammar for similar problems. I found that properties of `ObjectExpression` lack the `type` property (with value set to `"Property"`). I’ll fix that shortly.

atavakoli:
@dmajda thanks!

PR #440, opened 2016-08-24
Build LogicalExpression instead of BinaryExpression nodes as per ES5 spec

atavakoli:
The `buildLogicalExpression` function was defined but not used; specifically,
the `Logical(AND|OR)Expression(NoIn)?` rules were constructing `BinaryExpression`
nodes, but are now `LogicalExpression` nodes as per the [ESTree spec][1].

[1]: https://github.com/estree/estree/blob/2661f6647e875cac17a73f70f5e358507a900827/es5.md#logicalexpression

### DIFF
--- a/examples/javascript.pegjs
+++ b/examples/javascript.pegjs
@@ -855,12 +855,12 @@ BitwiseOROperator
 LogicalANDExpression
   = head:BitwiseORExpression
     tail:(__ LogicalANDOperator __ BitwiseORExpression)*
-    { return buildBinaryExpression(head, tail); }
+    { return buildLogicalExpression(head, tail); }
 
 LogicalANDExpressionNoIn
   = head:BitwiseORExpressionNoIn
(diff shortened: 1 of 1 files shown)


atavakoli:
@dmajda following your advice, I checked if there were any other functions that were unused, or possibly used to declare nodes against the standard, and nothing popped out at me, other than this one.

dmajda:
@atavakoli Thanks again! The patch looks good, I’d just maybe make the subject of the commit message little more specific (e.g. “JavaScript example: Use LogicalExpression nodes for "&&" and "||"”). With such change I’d be happy to merge.

atavakoli:
Sure thing @dmajda. I wasn't sure whether to prioritize making it <50 chars or making it descriptive.

dmajda:
@atavakoli Thanks. I tend to treat 50 chars for the subject as a soft limit which shouldn’t prevent descriptive messages. The hard one is 72.

Issue #429, opened 2016-06-21
Rewrite bin/pegjs using a command-line option parsing library


dmajda:
Right now, `bin/pegjs` parses options manually. This means some slightly advanced things like using `=` to separate option name and value or combining short options (`-abc` = `-a -b -c`) are not supported.

Let’s rewrite `bin/pegjs` using a proper command-line option parsing library which supports all these things out of the box. Possible candidates are [yargs](http://yargs.js.org/), [argp](https://www.npmjs.com/package/argp), or [cli](https://www.npmjs.com/package/cli).

seanich:
Do you mind if I take a crack at this? I'm fairly familiar with yargs so I'd like to use that if that's okay.

dmajda:
My requirements for any library used for the rewrite:

  * Full support for GNU-style option handling. See e.g. [here](http://www.catb.org/esr/writings/taoup/html/ch10s05.html), [here](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), or [here](http://docopt.org/) for a description.

  * Ability to customize error messages produced in cases like invalid option, missing option value, invalid option value, etc. This is important for two reasons:
    1. These messages need to be consistent with other messages produced by `bin/pegjs`.
    2. These messages are part of `bin/pegjs`’s interface which shouldn’t change if the underlying option-handling library changes in the future.

  * Ability to handle repeated options in different ways. For some options, specifying them repeatedly means the last value “wins” (e.g `pegjs -e foo -e bar` is the same as `pegjs -e bar`). For other options, all specified values are acted upon (e.g. `pegjs --plugin foo --plugin bar` loads both plugins).

  * It should be reasonably well written, documented, and maintained. It doesn’t necessarily have to be popular.

  * Using it should lead to code that is clean and readable.

@seanich I didn’t dive into yargs sufficiently to determine if is fulfills all the requirements (in fact, I’m not sure if any library does :-) But if you think so, feel free to go ahead with the rewrite if you still fell like it — just let me know. And sorry for not replying for so long — I was focusing at other things in the past weeks.

seanich:
I believe that yargs covers all of those requirements. I'll get to work on this and submit a PR this week. :)

dmajda:
I looked at [argp](https://github.com/gagle/node-argp).

> Full support for GNU-style option handling.

Not as robust as claimed. For example, consider this program:

```js
let argp = require("argp");

let argv = argp.createParser({ once: true })
  .body()
[...]

Issue #443, opened 2016-08-27
Update code style


dmajda:
There are few things I’d like to change on the style PEG.js code is written:

  * [x] **Use one `var`/`let`/`const` per variable.** The current code mostly uses something close to “one `var` per scope”, which is a style I picked up from [JavaScript: The Good Parts](http://shop.oreilly.com/product/9780596517748.do). The idea is to have just one `var` statement at the beginning of each function, which reflects the scope of declared variables. This will be less of an issue after switching to `let` & `const`. The current style also leads to somewhat irregular indentation.

  * [x] **Use only `//` comments.** The current style, which I carried it over from Java, uses both `//` and `/* ... */` comments. I used only `//` comments when working on [Juttle](https://github.com/juttle/juttle) and I liked the simplicity and consistency.

  * [x] **Use only double quotes for strings.** The current style is to use double quotes except when the content of the string is code, in which case single quotes are used. This is confusing and it’s better to just standardize on one kind of quotes. I chose double quotes because I’m used to single quotes delimiting a character (not string) literal in many languages (e.g. C, Java).

  * [x] **Avoid aligning object keys, `=` signs, etc.** Aligning leads to various inconsistencies and ugliness (e.g. in object literals with both single-line and multi-line values), is costly to maintain, and it tends to fill diffs with irrelevant changes. Again, this is something I carry over from Juttle.

I may add more things here over time. Ultimately, all these style rules should be [enforced by ESLint](https://github.com/pegjs/pegjs/issues/407).

Issue #442, opened 2016-08-25
Drop Node.js 0.10.x & 0.12.x support and update code accordingly


dmajda:
Node.js 0.10.x & 0.12.x is obsolete and its support can be dropped (its maintenance ends in Oct/Dec 2016, see also the [Twitter poll](https://twitter.com/peg_js/status/767628930253746176)). This allows us to modernize the code a bit — mainly use some ES6 features.

Should be done after #441.

Tasks:

  * [x] Drop Node 0.10.x & 0.12.x support
    * [x] Update `package.json`
    * [x] Update the “Compatibility” section in `README.md`
    * [x] Don’t test Node 0.10.x & 0.12.x in Travis CI
  * [x] Babelify all code that can be seen by the browser
    * [x] Pass the browser version through Babel
    * [x] Pass spec code through Babel before serving it to the browser
    * [x] Pass benchmark code through Babel before serving it to the browser
  * [x] Update code
    * [x] Replace `var` with `let` & `const`
    * [x] Replace some functions with arrow functions
    * [x] Replace `classes.subclass` with native classes
    * [x] ~~Replace suitable `for` loops with `for-of`~~
    * [x] ~~Replace `Array.prototype.forEach` with `for-of`~~
    * [x] Replace remaining functions in `utils` with native ones where possible

dmajda:
This is done now, but I’ll add two comments (both as notes-to-self and as an explanation for anyone interested):

  * **I decided to use `Array.prototype.forEach` in instead of `for-of` in the codebase.** Originally, I intended to use `for-of` because it seems more universal and more deeply integrated into the language (via [iteration protocols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols)). Then I prototyped the change and discovered that:

    * `Array.prototype.forEach` composes better with the rest of the code. For example, it is [possible to pass it a reference to a function which is used elsewhere](https://github.com/pegjs/pegjs/blob/b2a9dabcff19e339f32cb0dbfcffe7de15c7bcc9/lib/compiler/passes/generate-bytecode.js#L293). There is no need to spell out the actual call.

    * In at least one instance I needed an index associated with the iterated item. This is trivial with `Array.prototype.forEach` but impossible with `for-of`.

[...]

Issue #407, opened 2016-01-25
Use ESLint to enforce code style


dmajda:
After [switching from JSHint to ESLint](https://github.com/pegjs/pegjs/issues/406), use it to enforce coding conventions. More specifically, go through [ESLint’s rules](http://eslint.org/docs/rules/), enable all that make sense, and configure them to match how PEG.js code is written.

silkentrance:
Have a look at https://github.com/coldrye-es/eslint-config-coldrye for a starter. These rules are rather restrictive, so you might want to loosen them up a bit.

dmajda:
@silkentrance Thanks but I’ll go build the rule configuration myself.

dmajda:
This is done now. I ended up writing a detailed ESLint configuration and releasing it as a [separate project](https://github.com/dmajda/eslint-config-dmajda). This way it won’t be tied to PEG.js and I (or anyone else) will be able to use it in other projects.

pjmolina:
Very clean and elegant way to have ESLint settings apart.

PR #453, opened 2016-09-10
Rewrite bin/pegjs using yargs for options parsing

seanich:
This should resolve issue #429.

Sorry for the delay on this! Busy week at work.

I've tried to keep most of the behaviour identical to the previous version, however I did decide to change how the `--extra-options` and `--extra-options-file` arguments are handled.

Because of the way the extra options arguments were parsed in the previous version the last options specified would always take precedence (even if they came from an options file). In this new version the options specified as CLI arguments take precedence, then the options in `--extra-options`, then the options in `--extra-options-file`, and finally the default options. I think this is a more obvious behaviour.

### DIFF
--- a/bin/pegjs
+++ b/bin/pegjs
@@ -2,52 +2,21 @@
 
 "use strict";
 
-let fs      = require("fs");
-let path    = require("path");
-let peg     = require("../lib/peg");
+let fs = require("fs");
+let path = require("path");
+let yargs = require("yargs");
+
+let arrays = require("../lib/utils/arrays");
+let peg = require("../lib/peg");
 
 /* Helpers */
 
-function printVersion() {
-  console.log("PEG.js " + peg.VERSION);
-}
-
-function printHelp() {
-  console.log("Usage: pegjs [options] [--] [<input_file>]");
-  console.log("");
-  console.log("Options:");
-  console.log("      --allowed-start-rules <rules>  comma-separated list of rules the generated");
-  console.log("                                     parser will be allowed to start parsing");
-  console.log("                                     from (default: the first rule in the");
-  console.log("                                     grammar)");
-  console.log("      --cache                        make generated parser cache results");
-  console.log("  -d, --dependency <dependency>      use specified dependency (can be specified");
-  console.log("                                     multiple times)");
-  console.log("  -e, --export-var <variable>        name of a global variable into which the");
-  console.log("                                     parser object is assigned to when no module");
-  console.log("                                     loader is detected");
-  console.log("      --extra-options <options>      additional options (in JSON format) to pass");
-  console.log("                                     to peg.generate");
-  console.log("      --extra-options-file <file>    file with additional options (in JSON");
-  console.log("                                     format) to pass to peg.generate");
-  console.log("      --format <format>              format of the generated parser: amd,");
-  console.log("                                     commonjs, globals, umd (default: commonjs)");
-  console.log("  -h, --help                         print help and exit");
-  console.log("  -O, --optimize <goal>              select optimization for speed or size");
-  console.log("                                     (default: speed)");
-  console.log("  -o, --output <file>                output file");
-  console.log("      --plugin <plugin>              use a specified plugin (can be specified");
-  console.log("                                     multiple times)");
-  console.log("      --trace                        enable tracing in generated parser");
-  console.log("  -v, --version                      print version information and exit");
-}
-
-function exitSuccess() {
-  process.exit(0);
+function getVersionDescription() {
+  return "PEG.js " + peg.VERSION;
 }
 
 function exitFailure() {
-  process.exit(1);
+  process.exit(0);
 }
 
 function abort(message) {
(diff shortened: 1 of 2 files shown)


dmajda:
@seanich Thanks! I’ll have a look at the beginning of the next week.

dmajda:
@seanich Sorry I didn’t look yet, I wanted to finish work on #442 first (so that the code isn’t in the middle of a transition) which is taking a bit longer than I thought. Once I’m done there, I’ll focus on this PR.

dmajda:
I did a pass over this PR. As the first step I tested it without looking at the code much and I found multiple issues which need to be fixed.

Generic option parsing issues:

  * Arguments to short options can’t be specified without spaces (e.g. `-ofoo`). This should be supported as it is part of GNU-style option handling, one of the [requirements](https://github.com/pegjs/pegjs/issues/429#issuecomment-242047143).

  * Non-argument options can receive arguments (e.g. `-harg`, `—help=arg`). This should be an error.

  * Short options are also allowed as long options (e.g. `--O size`). This should be an error.

  * The message printed when an unknown option is given changed compared to the current state.

  * The message printed when an option is missing an argument changed compared to the current state. I’d be fine with changing the message to specify the option exactly (right now it talks about both the short and long form when they exist), but not much else.

  * The message printed when an option is passed an invalid argument (e.g. `--format invalid`) changed compared to the current state.

Issues tied to specific options:

  * The `--allowed-start-rules` options doesn’t work. For example, these two `bin/pegjs` invocations produce exactly the same parser, which they shouldn’t:

    ```console
    $ bin/pegjs examples/arithmetics.pegjs
    ```

    ```console
    $ bin/pegjs --allowed-start-rules Integer examples/arithmetics.pegjs
    ```

  * The error message printed when the `--dependency` option is used with an incompatible format is missing “.” at the end.

  * The `--export-var` option doesn’t check whether it is used with a compatible format. For example, this `bin/pegjs` invocation should lead to an error:

    ```console
    $ bin/pegjs --format commonjs --export-var parser examples/arithmetics.pegjs
    ```

  * The text printed when the `--help` option is used changed from the current state in undesirable way:

    * The usage line contains a full path to bin/pegjs:

      ```
      Usage: /Users/dmajda/Programming/PEG.js/pegjs/bin/pegjs [options] [--]
[<input_file>]
      ```

      It should contain just `pegjs`.

    * Options do not list their arguments.

    * Options list their types.

  If Yargs can’t be configured to avoid these changes, I’d rather see the help built and printed manually (like it its now).

These issues need to be fixed. Some are likely caused by Yargs itself and it may not be possible to avoid them. If that turns out to be the case, I’d unfortunately have to reject Yargs as a suitable command-line options parsing library.

As for `--extra-options` & `--extra-options-file`, I’d like to keep the current behavior. I find the semantics simple and consistent with how command-line options work in general (the ones specified later generally override those specified earlier).

If you continue working on the PR, please be aware that it got slightly out of sync with master as a consequence of changes done as part of resolving #442, so it will need to be re-synced. There are also upcoming changes to resolve #443 and #407, which will probably affect code in `bin/pegjs`. In any case, these will be formal/stylistic changes, not changes in functionality.

seanich:
> Arguments to short options can’t be specified without spaces (e.g. -ofoo). This should be supported as it is part of GNU-style option handling, one of the requirements.

This only applies to numeric arguments (eg `-O2`). `-ofoo` would correctly be interpreted as four combined mode-flag options which do not take arguments `-o -f -o -o`. 

> Non-argument options can receive arguments (e.g. -harg, —help=arg). This should be an error.

Once again `-harg` would be interpreted as `-h -a -r -g`. `--help=arg` is allowed since yargs tries to be fairly lenient about parsing. There may be some way to disable this, though.

> Short options are also allowed as long options (e.g. --O size). This should be an error.

Once again this is yargs being lenient. Perhaps we should look into a more strict option parsing library if this is a concern, though I would argue that this particular case is harmless.

> The message printed when an unknown option is given changed compared to the current state.

This is a message generated by yargs. As far as I know there is no way to change. Again, it might be worth looking into another library if this is an issue.

> The message printed when an option is missing an argument changed compared to the current state. I’d be fine with changing the message to specify the option exactly (right now it talks about both the short and long form when they exist), but not much else.

This won't be a problem to fix.

> The message printed when an option is passed an invalid argument (e.g. --format invalid) changed compared to the current state.

This is another message that is generated by yargs.

> The --allowed-start-rules options doesn’t work.

Oops, that is my mistake. I'll get that fixed.

> The error message printed when the --dependency option is used with an incompatible format is missing “.” at the end.

Thanks, I will fix that.

> The text printed when the --help option is used changed from the current state in undesirable way.

Yes, this is mostly unavoidable as it's generated by yargs without a lot of configuration options. It may in fact be best to manually generate the help text or to find a more configurable parsing library.

> The usage line contains a full path to bin/pegjs

The usage line prints the 0th positional parameter as it was passed to the program.

This seems to be an issue with what the shell is passing to node. The path is being expanded by the shell when invoked that way, but it will show expected results when invoked as `./bin/pegs`. This can be hard coded as `pegjs` again if that's your preference.

> As for --extra-options & --extra-options-file, I’d like to keep the current behavior. I find the semantics simple and consistent with how command-line options work in general (the ones specified later generally override those specified earlier).

Sure, no problem.

***

I'll look into some more option parsing libraries to see if I can find a better fit for this project. At a quick glance it seems like [cli](https://github.com/node-js-libs/cli) might provide the customization options we need, but I'll have to test it out.

dmajda:
> I'll look into some more option parsing libraries to see if I can find a better fit for this project.

@seanich Did you have a chance to do that? And more generally, are you still planning to work on this? If not, no problem — I’ll have a look at the rewrite myself then.

seanich:
I haven't had a chance to look into this unfortunately. I won't have much time in the next few weeks to work on this, so it might be best if you or someone else looked into it. Sorry about that!

dmajda:
OK, thanks.

PR #455, opened 2016-09-12
Add "use strict" and glob dev dependency for spec and benchmark servers

Mingun:


dmajda:
@Mingun Thanks for catching these!

Could you please split the change into two commits, one dealing with `"use strict"` and the other with `package.json`?

Mingun:
Done.

Issue #450, opened 2016-08-28
Don’t generate code in bytecode generator


dmajda:
In [various](https://github.com/pegjs/pegjs/blob/2a6a058c9c918dcbd615a7fff7464c1e73cce765/lib/compiler/passes/generate-bytecode.js#L204) [places](https://github.com/pegjs/pegjs/blob/2a6a058c9c918dcbd615a7fff7464c1e73cce765/lib/compiler/passes/generate-bytecode.js#L300) [in](https://github.com/pegjs/pegjs/blob/2a6a058c9c918dcbd615a7fff7464c1e73cce765/lib/compiler/passes/generate-bytecode.js#L537-L548) [the](https://github.com/pegjs/pegjs/blob/2a6a058c9c918dcbd615a7fff7464c1e73cce765/lib/compiler/passes/generate-bytecode.js#L565) [bytecode](https://github.com/pegjs/pegjs/blob/2a6a058c9c918dcbd615a7fff7464c1e73cce765/lib/compiler/passes/generate-bytecode.js#L601-L608) [generator](https://github.com/pegjs/pegjs/blob/2a6a058c9c918dcbd615a7fff7464c1e73cce765/lib/compiler/passes/generate-bytecode.js#L618) we actually generate code. This is conceptually wrong and it makes it harder to write non-JavaScript backends for PEG.js (it is not enough to replace the code generator, the bytecode generator needs to be modified too).

Let’s make sure the bytecode generator doesn’t generate a single piece of code.

Mingun:
This is implemented in 9b90fa1d8168878bcaef06be57e5d1f27489743c and can be closed

PR #459, opened 2016-09-17
Move all codegeneration from `generateBytecode` pass to `generateJs` pass

Mingun:


dmajda:
Nice! This is heading roughly in the direction I had in mind for #450.

First, let me give you some context. The original idea behind `consts` was that the bytecode should contain only numbers because that makes it possible to encode it efficiently when optimizing for size. Any other JavaScript value needs to be put on the side and referenced using an index. The important point is that **`consts` was intended to contain JavaScript values**.

The natural way to pass these values from the bytecode generator to the JavaScript generator was to pass them “as themselves”. This worked nicely for strings, regexps, etc. but it didn’t work for functions because [the result of calling `Function.prototype.toString` is implementation-dependent](http://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4.2). In practice, browsers at the time messed-up whitespace etc., while I wanted to preserve the code exactly. This led me to the simple solution of passing values around as JavaScript snippets. Still, **each snippet was just a stringified JavaScript value**.

Things got somewhat murkier after 5a4d04fa90471dd1ac1d7294c3083fcbdc732407 when this abstraction started breaking down. The snippets now represented JavaScript values only indirectly. This is what made me realize this approach is wrong and led me to file #450.

---

The way I see it, you try to do two things with this PR:

  1. Remodel the `consts` abstraction into something more fine-grained.
  2. Move actual code generation away from the bytecode generator.

I generally agree with both but these are largely orthogonal things. Could you please slice your changes so that there is one PR that does 1. (while still passing pieces of JavaScript around) and a follow-up PR which does 2. (a straightforward change)? This will make things much nicer.

As for replacing the `consts` abstraction, I think splitting `expectations` is perfectly fine and splitting `functions` as well. But I’m not so sure about splitting the rest into `literals` and `classes` and — more importantly — about the abstraction-level change in case of regexps.

The original idea here was to “lower” literals into JavaScript strings and classes into JavaScript regexps, then express matching at the bytecode level in terms of these. This allows some future optimizations, e.g. optimizing `start = "foo" / "bar"` into matching `/^foo|^bar/`. The disadvantage is of course that regexps are a JavaScript-specific construct.

I see 3 ways out:

  1. Stay with the current model, that is `consts` after splitting `expectations` and `functions` are JavaScript values. Pass these values around “as themselves” or in some other simple form.

    This would keep things simple but the bytecode generator would still be slightly JavaScript-specific as it would need to generate regexp patterns.

  2. Lift the level of abstraction and say that remaining `consts` are “abstract values” and pass them in some AST-like form. Something like:
    ```js
    {
      type: "string",
      value: "..."
    }
    ```
    ```js
    {
      type: "regexp",
      value: {
        type: "sequence",
        elements: [
          { type: "begin" }              // $
          { type: "class", parts: ... }  // [...]
        ]
      }
    }
    ```

    This would make the bytecode generator language-agnostic, but it would be somewhat complicated.

  3. Lift the level of abstraction and replace regexps with classes (what you did in this PR).

    This is reasonable technically but it would prevent future optimizations I talked about above.

Right now I’m in favor of keeping things simple and do 1., even at the cost of not being 100% language-agnostic and generating JavaScript regexps in the bytecode generator. But I’d like to hear your input.

---

Just as a side note, there are upcoming changes to resolve #443 and #407, which will undoubtedly affect code in this PR. If you continue working on this, you’ll probably need to re-sync with master periodically until these changes are finalized. Sorry about that. (I want to do all sweeping changes at the very beginning of the 0.11.0 cycle, before I look into more substantial things.)

Mingun:
Options 1 and 2 actually same. Just in the first case elements are already grouped in type, and in the second the type is attributed to each of them. Considering that with the abstract values it is all the same not necessary, it isn't clear, than option 2 is better than option 1. On the contrary, it seems to me, what is worse.

Not clearly, why in your option 2 sentence to present regexp in AST form. It will always have the same, in advance known structure, and therefore it is enough to store only its important parts. What I also made in the PR. Besides, use of regexp for comparison to a character class -- just pleasant feature of js. It is quite possible to do without any regexp -- rather simply to check entrance of the character in all described ranges (what I also do in the generator for pure C). Actually, the FSM produced by regexp does this namely.

> This is reasonable technically but it would prevent future optimizations I talked about above.

I don't see why the option 3 hinders with your "optimization". This "optimization" by definition is applicable only to js-generator -- so let and is executed at a code generation stage. Besides, I would call it rather deoptimization. For what reason changeover of one FSM by another shall change something? 

Besides, even if such optimization really will be quicker, the option of its application most likely will be negligible. For example, at what length to advance a parse position? Both strings contain 3 characters in your example, on in reality the identical length of literals will be rather an exception, than the rule. Also, the alternatives containing only literals most likely are rare. More really to meet alternatives of the sequences, and each alternative will have own action. The bigger scoring can be reached from [other][1] [optimizations][2].

------
> The way I see it, you try to do two things with this PR:

I split chages later

[1]: https://github.com/pegjs/pegjs/pull/399
[2]: https://github.com/pegjs/pegjs/pull/400

Mingun:
Also I thought and noted that separation of `classes` in a separate array nevertheless doesn't hinder with those optimization about which you spoke. As it will become at the level of AST, it will be just necessary to expand types of constants in `classes` array: except `class` shall there `choice` will appear, for example:
```js
{
  type: "choice",
  value: ["foo", "bar"]
}
```
JS code generator then generate regexp `/^foo|^bar/` from this definition.

dmajda:
> Options 1 and 2 actually same.

Not really. In the first case, the values would be *JavaScript values*, passed around in JavaScript-specific format (e.g. a regexp literal). In the second case, they would be *abstract values*, passed around in language-agnostic format.

In the first case, the bytecode generator and any code generator (even a non-JavaScript one) would need to contain a bit of JavaScript-specific logic, mostly to encode and parse regexps. In the second case, this would not be the case (at the cost of higher complexity).

> Not clearly, why in your option 2 sentence to present regexp in AST form. It will always have the same, in advance known structure, and therefore it is enough to store only its important parts. Besides, use of regexp for comparison to a character class -- just pleasant feature of js. 

In my world view, regexps are primitives of the abstract machine driven by the bytecode, used by `MATCH_REGEXP`. I don’t want the code generator to make (too much) assumptions about what kind of regexps it will receive. The fact that in the current implementation the regexps only contain character classes is incidental.

> I don't see why the option 3 hinders with your "optimization". This "optimization" by definition is applicable only to js-generator -- so let and is executed at a code generation stage.

Doing this optimization in the code generator would be too late. My idea was to optimize this at the AST level. In the `"foo" / "bar"` example I used, the optimization would replace the resulting `choice` node with two `literal` children with a `regexp` node (which would be later compiled into `MATCH_REGEXP`). 

Of course, this idea assumes that regexps are primitives of the abstract machine, as I discuss above.

> For what reason changeover of one FSM by another shall change something?

The underlying assumption is that regexp matching is optimized heavily by the JavaScript engine.

> For example, at what length to advance a parse position? Both strings contain 3 characters in your example, on in reality the identical length of literals will be rather an exception, than the rule.

Matching can be implemented using `RegExp.prototype.exec` which returns the matching string as part of the match result.

> Also, the alternatives containing only literals most likely are rare.

They are actually quite common in the lexical portion of grammars. Moreover, the optimization can potentially optimize much more — for example, the whole `number` rule in the JSON grammar example could be in principle compiled into one big regexp.

-----

I thought about this a bit more. I think I’d like to keep regexps as a primitive in the abstract machine and not replace them by a class. This will keep the machine JavaScript-specific, but everything else will be simpler.

Obviously, this is not ideal for generators into languages other than JavaScript, but to be honest, these are mostly an afterthought. My policy (now formulated explicitly for the first time here) is that I don’t want to complicate creating them, but only when that doesn’t significantly hinder other goals. (As a side note, there are other things that would need to change to make creating them fully supported, like making label reserved word detection configurable because each language has different reserved words. In any case, this is not something I want to invest effort in right now.)

I don’t think all this invalidates #450 — moving (almost) all code generation out of the bytecode generator is still a worthy goal.

Technically, this translates into option 1 with `expectations`, `functions` and `consts` arrays, where `consts` would contain JavaScript values as themselves (the simplest solution). If you could rework your PRs in that direction, it would be great.

Mingun:
@futagoza are you interesting to merge this work? Should I spend time for rebase?

futagoza:
If you have time, that would be great.

Mingun:
This ready to merge. Also I can add additional commit on top of this, that replaces opcode `PUSH` with `PUSH_EMPTY_STRING` because now it only used to push empty string (see https://github.com/Mingun/pegjs/commit/a86c58e762aae6ba61b93ff5828360d4fc9b1e49)

futagoza:
If `PUSH` is only used to push an empty string, then yes, it is a good idea to change it to `PUSH_EMPTY_STRING`, that cause's less confusion when looking at the bytecode generator.

Also, it would be nice if you could update the [typings](https://github.com/Mingun/pegjs/blob/bc-generator-cleanup/lib/typings/api.d.ts#L61) 😄

Mingun:
Unfortunately, I don't know typescript and I don't represent where to place the required declarations. So I can't update typings, but I believe what has to be approximately so:
```ts
interface ClassConst {

    value: (string | string[])[];
    inverted: boolean;
    ignoreCase: boolean;

}

type ExpectedConst
    = ExpectedAnyConst
    | ExpectedClassConst
    | ExpectedLiteralConst
    | ExpectedRuleConst;

interface ExpectedAnyConst {

    type: "any";

}

interface ExpectedClassConst extends ClassConst {

    type: "class";

}

interface ExpectedLiteralConst {

    type: "literal";
    value: string;
    ignoreCase: boolean;

}

interface ExpectedRuleConst {

    type: "rule";
    value: string;

}

interface FunctionConst {

    predicate: boolean;
    params: string[];
    code: string;

}


interface Grammar extends INode {
...
    // Added by Bytecode generator

    literals?: string[];
    classes?: ClassConst[];
    expectations?: ExpectedConst[];
    functions?: FunctionConst[];
...
}
```

futagoza:
It's okay, I'll add they typings later

PR #471, opened 2016-11-08
allow whitespace on beginning and end of expressions

tonylukasavage:


dmajda:
Thanks for the contribution but I’d rather keep the grammar simple and the `Expression` rule symmetrical to the `Term` rule.

PR #475, opened 2016-11-27
Add some useful debug information to some exceptions

Mingun:


spec/unit/compiler/passes/helpers.js
```diff
@@ -68,7 +68,7 @@ beforeEach(function() {
           this.message = () =>
             "Expected the pass not to report an error "
               + "for grammar " + jasmine.pp(grammar) + ", "
-              + "but it did.";
+              + "but it did with message " + jasmine.pp(e.message);
```
dmajda: Please replace “but it did with message” with more grammatically correct “but it reported one with message”. Also, there was a comma at the end of the message.

lib/compiler/passes/generate-js.js
```diff
@@ -383,7 +383,7 @@ function generateJS(ast, options) {
       "          break;",
       "",
       "        default:",
-      "          throw new Error(\"Invalid opcode: \" + bc[ip] + \".\");",
+      "          throw new Error(\"Invalid opcode #\" + ip + \": \" + bc[ip] + \".\");",
```
dmajda: I’m curious: you apparently missed the opcode position in the message but you didn’t miss an indication in bytecode of *which rule* the invalid opcode was? I’d rather add both, e.g. “Invalid opcode: 77 (rule: 5, position: 6).”.

lib/compiler/passes/generate-js.js
```diff
@@ -473,7 +473,7 @@ function generateJS(ast, options) {
 
           if (thenSp !== elseSp) {
             throw new Error(
-              "Branches of a condition must move the stack pointer in the same way."
+              "Branches of a condition must move the stack pointer in the same way (then=" + thenSp + ", else=" + elseSp + ")."
```
dmajda: I’d format the message like this for more readability: “Branches of a condition must move the stack pointer in the same way (before: 5, after then: 6, after else: 4).”. Or maybe even like this: “Branches of a condition can't move the stack pointer differently (before: 5, after then: 6, after else: 4).”.

Also please split the concatenation over multiple lines.

lib/compiler/passes/generate-js.js
```diff
@@ -499,7 +499,7 @@ function generateJS(ast, options) {
         ip += bodyLength;
 
         if (bodySp !== baseSp) {
-          throw new Error("Body of a loop can't move the stack pointer.");
+          throw new Error("Body of a loop can't move the stack pointer (base=" + baseSp + ", body=" + bodySp + ").");
```
dmajda: I’d format the message like this for more readability: “Body of a loop can't move the stack pointer (before: 5, after: 6).”.

Also please split the concatenation over multiple lines.

lib/compiler/passes/generate-js.js
```diff
@@ -705,7 +705,7 @@ function generateJS(ast, options) {
             break;
 
           default:
-            throw new Error("Invalid opcode: " + bc[ip] + ".");
+            throw new Error("Invalid opcode #" + ip + ": " + bc[ip] + ".");
```
dmajda: Like above, I’d add both the position and the rule into the message: “Invalid opcode: 77 (rule: Expression, position: 6).”.

dmajda:
As a side note, the Travis CI [build failure](https://travis-ci.org/pegjs/pegjs/builds/179220657) is unrelated to this change and I [fixed it](https://github.com/pegjs/pegjs/commit/a733beb933d67b725c4972c97b8e5929792224bd) yesterday. Rebasing on top of current master should result in a clean build.

Mingun:
Any comments?

Mingun:
@futagoza?

Mingun:
Now this done and can be merged.

futagoza:
I prefer `fn` over `func`, but oh well, doesn't matter 😄

PR #480, opened 2016-12-07
Fix LabelledStatement rule to use same spelling as node

federicobond:


dmajda:
@federicobond As stated in the header comment, the JavaScript example grammar is based on a [grammar](https://www.ecma-international.org/ecma-262/5.1/#sec-A) from ECMA-262, 5.1 Edition. That grammar spells the rule name as `LabelledStatement`. I think it’s best to keep that spelling in order to maintain correspondence between rules in both grammars.

Issue #423, opened 2016-04-07
Support exporting parsers as ES2015 modules


fatfisz:
Currently the `buildParser` method returns either an object or an IIFE. This prevents seamless ES6-style imports.

There is a plugin that allows importing parsers built from grammar files, `rollup-plugin-pegjs`. It currently can't do much more than either `export default generatedParser;` or `module.exports = generatedParser;`.

Despite the claims in the `rollup-plugin-pegjs` README that

``` js
import { parse } from "grammar.pegjs";
```

is possible, I've been unable to do so and have to rely on this:

``` js
import parser from "grammar.pegjs";
const { parse } = parser;
```

So taking that into consideration, would it be ok if there was another option for `output`: `source-es6` or `source-es6-module` that would have IIFE stripped and would have ES6-style exports instead of returning an object?

I could do the PR if the idea is accepted.

dmajda:
The way I see it, there are two questions in play here:

  1. What is the responsibility of the `buildParser` function?
  2. What exactly should be exported?

As for 1, in the current architecture `buildParser` returns either the parser object or source code which produces it when evaluated. This invariant is [encoded into the compiler](https://github.com/pegjs/pegjs/blob/0c39f1cf86d73c85a1f832e6cfddf4176f9e58bc/lib/compiler.js#L59..L62). The function doesn’t deal with the module system at all (this is [done  in `bin/pegjs`](https://github.com/pegjs/pegjs/blob/0c39f1cf86d73c85a1f832e6cfddf4176f9e58bc/bin/pegjs#L270)).

This design is simple and understandable, but your request made me realize that it’s probably not the most practical one. The code which handles exporting currently has to be replicated in many tools which use PEG.js API (e.g. [`gulp-peg`](https://github.com/lazutkin/gulp-peg/blob/d76f75549ca88e3e1f58b0418dc55c6acdf5a374/src/index.coffee#L10) or [`grunt-peg`](https://github.com/dvberkel/grunt-peg/blob/7438cc4cba68fe3fd256a8076019d7f3876e2416/tasks/peg.js#L39)). It would be more useful to put export handling into `buildParser` itself (either by allowing more `output` values or by adding specialized option(s) valid when `output` is set to `"source"`). I’ll have a look at this as part of #362 (I just added an item into a checklist there).

[...]

PR #481, opened 2016-12-10
Add support for ES modules

fatfisz:
Related to #423.

A couple of notes:
- I noticed that the only place where the options are being validated is in the `bin/pegjs` script, shouldn't it be in the `generate` method instead?
- I couldn't find any existing tests for different output formats, are there any?

This does not yet contain changes to README, which I will add later if the code is good first.

### DIFF
--- a/bin/pegjs
+++ b/bin/pegjs
@@ -30,8 +30,8 @@ function printHelp() {
   console.log("                                     to peg.generate");
   console.log("      --extra-options-file <file>    file with additional options (in JSON");
   console.log("                                     format) to pass to peg.generate");
-  console.log("      --format <format>              format of the generated parser: amd,");
-  console.log("                                     commonjs, globals, umd (default: commonjs)");
+  console.log("      --format <format>              format of the generated parser: amd, commonjs");
+  console.log("                                     es, globals, umd (default: commonjs)");
   console.log("  -h, --help                         print help and exit");
   console.log("  -O, --optimize <goal>              select optimization for speed or size");
   console.log("                                     (default: speed)");
--- a/bin/pegjs
+++ b/bin/pegjs
@@ -116,6 +116,9 @@ let options = {
   trace: false
 };
 
+const MODULE_FORMATS = ["amd", "commonjs", "es", "globals", "umd"];
+const MODULE_FORMATS_WITH_DEPS = ["amd", "commonjs", "es", "umd"];
+
 while (args.length > 0 && isOption(args[0])) {
   let json, id, mod;
 
@@ -183,8 +186,8 @@ while (args.length > 0 && isOption(args[0])) {
       if (args.length === 0) {
         abort("Missing parameter of the --format option.");
       }
-      if (args[0] !== "amd" && args[0] !== "commonjs" && args[0] !== "es" && args[0] !== "globals" && args[0] !== "umd") {
-        abort("Module format must be one of \"amd\", \"commonjs\", \"es\", \"globals\", and \"umd\".");
+      if (MODULE_FORMATS.indexOf(args[0]) === -1) {
+        abort("Module format must be one of " + MODULE_FORMATS.map(format => `"${format}"`).join(", ") + ".");
       }
       options.format = args[0];
       break;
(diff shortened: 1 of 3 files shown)


bin/pegjs
```diff
@@ -183,8 +183,8 @@ while (args.length > 0 && isOption(args[0])) {
       if (args.length === 0) {
         abort("Missing parameter of the --format option.");
       }
-      if (args[0] !== "amd" && args[0] !== "commonjs" && args[0] !== "globals" && args[0] !== "umd") {
-        abort("Module format must be one of \"amd\", \"commonjs\", \"globals\", and \"umd\".");
+      if (args[0] !== "amd" && args[0] !== "commonjs" && args[0] !== "esm" && args[0] !== "globals" && args[0] !== "umd") {
+        abort("Module format must be one of \"amd\", \"commonjs\", \"esm\", \"globals\", and \"umd\".");
```
dmajda: I think it would be slightly better to identify the format as `es2015` — it is more readily understandable. But if most common tools that work with module formats use `esm`, then it would be probably better to keep it (I’d like to see examples of such tools then).

The condition also got a bit too long — maybe extract the list of formats into a const  named e.g. `MODULE_FORMATS` and rewrite the test as `MODULE_FORMATS.indexOf(args[0]) === -1`? Up to you. Best done as a separate commit before the addition itself.

bin/pegjs
```diff
@@ -253,7 +253,7 @@ while (args.length > 0 && isOption(args[0])) {
 }
 
 if (Object.keys(options.dependencies).length > 0) {
-  if (options.format !== "amd" && options.format !== "commonjs" && options.format !== "umd") {
+  if (options.format !== "amd" && options.format !== "commonjs" && options.format !== "esm" && options.format !== "umd") {
```
dmajda: Same as above (`es2015`, maybe long condition → `MODULE_FORMATS_WITH_DEPS`).

dmajda:
@fatfisz Thanks! I just reviewed.

> - I noticed that the only place where the options are being validated is in the `bin/pegjs` script, shouldn't it be in the `generate` method instead?

My philosophy is not doing parameter validation in functions/methods when I work in dynamic languages because it would pretty much lead to duplication of the (absent) type system if done consistently. On the other hand, a program shouldn’t crash or produce an exception because of bad user input and it should produce appropriate error messages for given interface, which logically leads to validation in `bin/pegjs`.

> - I couldn't find any existing tests for different output formats, are there any?

No. I think the right way to test this would be using integration tests — that is, compile a parser, load it in a suitable environment for given module format, and verify that the exports are there and they work.

Setting this up is obviously a bit of work, so I chose not to do it given that all module-related code is self-contained and unlikely to break without directly touching it. But I would accept such tests as a contribution.

For this PR this means relying on manual testing.

fatfisz:
About the name: Babel uses ["es2015"](https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/), Rollup uses ["es"](https://github.com/rollup/rollup/wiki/JavaScript-API), Webpack doesn't have the target yet, though the suggested name at the moment is ["module"](https://github.com/webpack/webpack/issues/2933).

Also, I suggested "esm" based on what I've seen in the discussions about what should be the extension for ES modules in Node.js. Though "es" and "jsm" was also listed there.

Out of those my personal preference is "es" - the same module system will be there in ES2017, ES2018, etc., so sticking to the version in this case seems a bit pointless, but I could be convinced otherwise.

dmajda:
I think you convinced me. Having a year in the module format name may be important for projects that deal with multiple ECMAScript versions (like Babel or TypeScript), but not so much here. And it’s unlikely that the module system will ever change in a way that would justify using the year for versioning.

Let’s go with `es`.

arcanis:
Any news about this? I currently have some troubles making Rollup understand that the parser actually does export a `parse` symbol, and I think this PR could help.

tilgovi:
This branch works great for me!

futagoza:
@fatfisz Github is not letting me merge thi due to conflicts in _README.md_, so if you can just fix those, I'm happy to merge it straight away

fatfisz:
Sure, will see if I can get it done today!

fatfisz:
Rebased + added a missing piece of info to `bin/pegjs`.

futagoza:
@fatfisz Thanks, I've merged it now 😄

allnulled:
Hi folks, I suggest to add this feature information into the real official documentation of the project.
Thanks for your time, folks.

PR #490, opened 2017-02-15
CSS example: parse decimal form of nums correctly

chearon:
Hello, thanks for making this. I've been using a variant of the example file `css.pegjs` for something I've been working on, and I think I may have ran into a small mistake in it.

The `nums` rule was trying whole numbers before trying decimals, so if you had a rule like `nums+`, it would parse `0.02` as 2 `num` tokens - 0 and .02. I think this is the right fix for it.


futagoza:
@chearon Good catch, thanks

PR #492, opened 2017-02-23
Do not indent backtick quoted strings in code blocks

alanmimms:
I found the functions `indent2()`, `indent6()` and `indent10()` will insert indentation in code even if it has multi-line backtick strings in it. To avoid this, I changed the regex used in these functions to match their source strings. It might be a teeny bit faster to move the initialization of `lineMatchRE` I created outside the `generateJS()` function since its regex will then only get compiled one time.


lib/compiler/passes/generate-js.js
```diff
@@ -6,9 +6,11 @@ let op = require("../opcodes");
 
 // Generates parser JavaScript code.
 function generateJS(ast, options) {
-  // These only indent non-empty lines to avoid trailing whitespace.
-  function indent2(code)  { return code.replace(/^(.+)$/gm, "  $1");         }
-  function indent10(code) { return code.replace(/^(.+)$/gm, "          $1"); }
+  /* These only indent non-empty lines to avoid trailing whitespace. */
+  var lineMatchRE = /^([^`\r\n]+?(?:`[^`]*?`[^\r\n]*?)?)$/gm;
```
futagoza: you might want to change this to a `const` variable to pass eslint

alanmimms:
Wow! I thought about doing that, but then thought "This code has to run on old versions of Node" and went with `var`. Doesn't this code need to run on old versions of Node? That's why I didn't do it.

futagoza:
PEG.js dropped support for Node.js 0.10.x & 0.12.x

The minimum required version is now Node.js 4, although Babel is used through
babelify to support ES5 browsers.

futagoza:
@alanmimms I;m planning to squash and merge this, is that fine with you?

alanmimms:
Fine with me. Thanks.

YemSalat:
@alanmimms @futagoza tbh, its a bit scary to see a significant change in regex in `generateJS` w/o tests. I think those should probably be added in a separate pr?

futagoza:
@YemSalat Thanks for reminding me, before merging I was trying to remember why I delayed this PR 😆 

@alanmimms You also willing to do the tests? If not, just tell me so I can put in on my todo list 😄

PR #496, opened 2017-03-31
A lightweight, but mostly compatible version of JS grammar which is easier to grok.

mikeaustin:


YemSalat:
That is quite a simplified grammar, but I think the main problem here is that it is not JS anymore, and I am guessing the idea of JS grammar example is to be able to take it and see it parse actual JS code.

Some notes on adjustments you made:
> * Removed automatic semicolon insertion. Everyone uses semicolons. Simplifies
     the language and also eliminates dangling curly brace issues.

I think you are not correct about everyone using semicolons. E.g. check out: https://github.com/feross/standard

> * Removed break, continue, debugger, delete, do, instanceof, in, void, while
     and with rules. Functionality can be implemented with functions. Using
     forEach(), map(), etc. also fix closure variable capturing.
> * Removed prefix and postfix operators ++ and --. Modern languages don't need
     these. You can use i += 1, or iterator.moveNext() instead of ++iterator.
> * Removed regular expression rules. Use a string instead of special syntax.
> * Removed +=, -=, *=, /=, %= operators. Handy, but complicate the language.
Maybe I'll add them back if they're sorely missed.

I think all these are used in JS way too often to remove them.

> * Removed === and !== operators. A transpiler would convert all == to ===.

That would break code which checks for both null/undefined using double equals

mikeaustin:
Yeah, maybe a pull is not the right thing to do. I'm just toying with making a dialect of JavaScript and thought it would be useful for others to see a stripped down version. It's not JS really :)

I'll just keep working on it and keep it separate. Maybe a link to it is a better option if useful.

I just added range and tuple syntax (1..5, 'a'..'z') and fat-arrow functions. We'll see what happens.

futagoza:
The example grammars illustrate the different features of PEG.js and how to use them, as well as provide examples on how to use JavaScript in actions to alter, evaluate and return data from rules.

As far as I can tell @dmajda wont accept PR's that will add new grammar files, but you can add a link to you grammar file here: https://github.com/pegjs/pegjs/wiki/Useful-Grammars

PR #497, opened 2017-04-06
Auto Dependency Updates ...

lguzzon:


lguzzon:
Waiting patiently ...

futagoza:
@lguzzon Thanks for this, but I've decided to keep most of the dependencies locked untill after 0.11, at which point I will start a restructure and rewrite of the development files, which will change dependencies.

lguzzon:
Never mind ;)

PR #509, opened 2017-06-24
Simplify bytecode: remove unnecessary opcode.

Mingun:
Now generated sources looks little better :)

### DIFF
--- a/lib/compiler/passes/generate-bytecode.js
+++ b/lib/compiler/passes/generate-bytecode.js
@@ -407,11 +407,10 @@ function generateBytecode(ast) {
               [op.LOAD_SAVED_POS, node.elements.length],
               buildCall(
                 functionIndex,
-                node.elements.length,
+                node.elements.length + 1,
                 context.env,
                 context.sp
-              ),
-              [op.NIP]
+              )
             );
           } else {
             return buildSequence([op.WRAP, node.elements.length], [op.NIP]);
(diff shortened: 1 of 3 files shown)


futagoza:
As I'm still not familar with the bytecode generator, so can you please explain some stuff before I decided to merge this PR.

From my understanding `buildCall` returns an array that is concated by `buildSequance`, so how does removing `[op.NIP]` from the call to `buildSequance` and adding `node.elements.length + 1` to the call to `buildCall` affect the built array returned by `buildSequance`?

Mingun:
`buildCall` constructs opcode `CALL` with necessary parameters. `buildSequence` just concatenate two or more arrays of opcodes, in this case `CALL` opcode with `NIP` opcode. According to description of [`CALL`][1] and [`NIP`][2] stack modified in such way:
```js
// CALL
value = consts[f](stack[p1], ..., stack[pN]);
stack.pop(n);
stack.push(value);
// NIP
value = stack.pop();
stack.pop();
stack.push(value);

// After simplifications code are:
value = consts[f](stack[p1], ..., stack[pN]);
stack.pop(n);
stack.pop();
stack.push(value);
// which is also CALL opcode but with one more pop'ed values
```

[1]: https://github.com/pegjs/pegjs/blob/master/lib/compiler/passes/generate-bytecode.js#L167-L171
[2]: https://github.com/pegjs/pegjs/blob/master/lib/compiler/passes/generate-bytecode.js#L52-L56

PR #511, opened 2017-06-24
Add ability to extract comments from the grammar

Mingun:
All comments stored in the `comments` property of the `grammar` node.
This property is object with mapping start offset of comment to comment object, that looks like:

```js
{
  text: 'text in the comment, just after // or /* and before */',
  multiline: true|false,// true for /**/ comments, false for // comments
  location: location()
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,25 @@ object to `peg.generate`. The following options are supported:
   * `plugins` — plugins to use
   * `trace` — makes the parser trace its progress (default: `false`)
 
+Also you can supply boolean option `extractComments` (default: `false`).
+When set to `true`, parser will be collect all comments in the grammar to the object
+in the `comments` property in the `grammar` AST node. This key contains mapping from
+offset position of start location of the comment (i.e. `//` or `/*`) to comment object
+itself. Comment object has following structure:
+
+  ```js
+  {
+    text: 'all text between /* or */, or // and end of line',
+    multiline: true|false,
+    location: location()
+  }
+  ```
+
+When set to `false`, `comments` will be set to `null`.
+
+This option not impact to the generated parser but only to grammar AST, which
+can be used by plugins.
+
 Using the Parser
 ----------------
 
--- a/lib/parser/ast.js
+++ b/lib/parser/ast.js
@@ -15,11 +15,12 @@ exports.Node = Node;
 class Grammar extends Node {
 
     // Creates a new AST
-    constructor( initializer, rules, location ) {
+    constructor( initializer, rules, comments, location ) {
 
         super( "grammar", location );
 
         this.initializer = initializer;
+        this.comments = comments;
         this.rules = rules;
 
     }
--- a/lib/parser/index.js
+++ b/lib/parser/index.js
@@ -237,6 +237,7 @@ function peg$parse(input, options) {
         return new ast.Grammar(
           extractOptional(initializer, 0),
           extractList(rules, 0),
+          comments,
           location()
         );
       };
@@ -311,25 +312,31 @@ function peg$parse(input, options) {
   var peg$f11 = function(operator, code) {
         return createNode( OPS_TO_SEMANTIC_PREDICATE_TYPES[operator], { code: code } );
       };
-  var peg$f12 = function(name) { return [name, location()]; };
-  var peg$f13 = function(head, tail) { return head + tail.join(""); };
-  var peg$f14 = function(sequence) { return sequence; };
-  var peg$f15 = function(value, ignoreCase) {
+  var peg$f12 = function(comment) {
+      return addComment(comment, true);
+    };
+  var peg$f13 = function(comment) {
+      return addComment(comment, false);
+    };
+  var peg$f14 = function(name) { return [name, location()]; };
+  var peg$f15 = function(head, tail) { return head + tail.join(""); };
+  var peg$f16 = function(sequence) { return sequence; };
+  var peg$f17 = function(value, ignoreCase) {
         return createNode( "literal", {
           value: value,
           ignoreCase: ignoreCase !== null,
         } );
       };
-  var peg$f16 = function(chars) { return chars.join(""); };
-  var peg$f17 = function() { return text(); };
-  var peg$f18 = function(inverted, parts, ignoreCase) {
+  var peg$f18 = function(chars) { return chars.join(""); };
+  var peg$f19 = function() { return text(); };
+  var peg$f20 = function(inverted, parts, ignoreCase) {
         return createNode( "class", {
           parts: parts.filter(part => part !== ""),
           inverted: inverted !== null,
           ignoreCase: ignoreCase !== null,
         } );
       };
-  var peg$f19 = function(begin, end) {
+  var peg$f21 = function(begin, end) {
         if (begin.charCodeAt(0) > end.charCodeAt(0)) {
           error(
             "Invalid character range: " + text() + "."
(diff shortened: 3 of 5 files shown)


futagoza:
What would this be useful for?

Mingun:
For example for creation of the full-featured editor for pegjs grammars. Assigning comments to the AST nodes is the task for separate pass which won't be a pegjs core part however for the work it demands or own grammar (why then a plug-in if everything needs to be rewritten?), or to have support from pegjs core. This PR provides such support.

futagoza:
For editor support wouldn't it be better to do this:
```pegjs
Rule
  = comment:Comment __
    name:IdentifierName __
    displayName:(StringLiteral __)?
    "=" __
    expression:Expression EOS
    {
      return {
        type: "rule",
        name: name,
        expression: displayName !== null
          ? {
              type: "named",
              name: displayName[0],
              expression: expression,
              location: location()
            }
          : expression,
        location: location(),
        comment: comment
      };
}

MultiLineComment
  = "/*" comment:$(!"*/" SourceCharacter)* "*/" {
    return { type: "MultiLineComment", value: comment, location: location() };
  }

MultiLineCommentNoLineTerminator
  = "/*" comment:$(!("*/" / LineTerminator) SourceCharacter)* "*/" {
    return { type: "MultiLineComment", value: comment, location: location() };
}

SingleLineComment
  = "//" comment:$(!LineTerminator SourceCharacter)* {
    return { type: "SingleLineComment", value: comment, location: location() };
}
```

Mingun:
You solution not cover all possible variants of comments. For example, that I am using:
```js
// Do not forget about grammar-level comments

// One
// big
// rule
// comment
rule1 = ...;

rule2// rule comment
  // first alternative comment
  = ...
  // second alternative comment
  / ...
  ...
  ;

// or such:
rule3
  = 'a' /* sequence part1 comment */
    'b' /* sequence part2 comment */
    ...
  ;
```
My solution covers such cases. Other things being equal I prefer to have more information, than it is less.

Concerning impact -- I don't think that simple addition to a hash table somehow will worsen it (hash is used instead array because comment rules may be parsed multiply times in one position thanks to backtracking). The main work has to the parse itself.

futagoza:
Your right, some people prefer using line comments as block comments, so this would be better (applied to my above example):

```pegjs
Rule
    = comment:RuleComment __
    ...

RuleComment
    = head:SingleLineComment tail:(ws SingleLineComment)* {
        return tail.length ? {
            type: "CommentBlock",
            comments: buildList(head, tail, 1),
            location: location()
        } : head;
     }
    / MultiLineComment

ws = (WhiteSpace / LineTerminatorSequence)*
```

I'm not sure that having the parser also gather comments that are in other places (between the rule's name and assignment operator, between expressions, between literals, etc) is usefull as the only thing an editor can show a comment for is a refrence to a rule. I'm not saying having comments in those locations is bad (I also keep comments in random locations like that to keep track of stuff and leave notes to my self), it's just that an editor wouldn't really do anything with these comments, at least not one that I know of.

Also for speed impact, your right, the hash table does not add any impact to the parser's performance, but the lookup calls for the location do, and when creating a parser that will be used for an editor, I'd say the quicker the responce, the better, so ensuring less calls to the location function is better.

The idea though isn't a bad one, it's just this should be either an opt-in behaviour, or one that is added via a plugin. I'm brainstroming idea's on where to take PEG.js after 1.0, and one Idea that I'm juggeling around right now is adding another generator that would generate class based parsers, which would allow the generated parser's to be extendable. If I do go in this direction, then generating partial parser's that can be used to extend other parser's would be a good direction for this idea to take.

Mingun:
> as the only thing an editor can show a comment for is a refrence to a rule.

Not only. For example, it is possible to fold the several lines of comments. It is obvious that it is possible to fold not only rule comments.

What - the decision with the partial implementation of an feature wouldn't be selected, in of course total it will lead to the fact that the thing which, apparently, can be made, made is impossible. I am very much baffled by similar finds.

I thought much on how to extract comments from the file. Finally suggested solution the simplest.

> but the lookup calls for the location do
Costs for reading the hash table will be only once - a code it will be analyzed and place links in the AST nodes to the comments corresponding to them. But, as I already told, it will do in the separate pass.

futagoza:
> What - the decision with the partial implementation of an feature wouldn't be selected, in of course total it will lead to the fact that the thing which, apparently, can be made, made is impossible. I am very much baffled by similar finds.

I'm sorry but due to your choice of words I don't quite understand, but I'm presuming your saying that it's not going to be a feature used because it wont be posiible to do what the original parser was generated to do due to the partial parser changing it's API, is that what your saying? If so, isn't that the point of an extendable parser? To more easily prototype and add feaures the orignal wasn't meant to do, or restrict what it does.

> Costs for reading the hash table will be only once

Like I said above, the hash table doesn't affect the generated parser's performance, but calling `location()` alot can affect performance, especially right now as multi file grammars haven't been implemented yet (using the currently generated parser for PEG.js as an example):
```javascript
// location() returns the result of this:
peg$computeLocation(peg$savedPos, peg$currPos);

// peg$computeLocation() then calls these:
var startPosDetails = peg$computePosDetails(startPos); // peg$savedPos
var endPosDetails = peg$computePosDetails(endPos); // peg$currPos

// location @ https://github.com/pegjs/pegjs/blob/master/lib/parser.js#L505
// peg$computeLocation @ https://github.com/pegjs/pegjs/blob/master/lib/parser.js#L584
// peg$computePosDetails @ https://github.com/pegjs/pegjs/blob/master/lib/parser.js#L549
```
As you can see from the source, `peg$computePosDetails` will pull the details from `peg$posDetailsCache` array, but only if details have been cached. As this PR will be adding the feature to extract comments with the location details, it means `peg$computePosDetails()` will be called for each comment twice, but since nothing else in the parser is looking up location details for the index's where the comments start and end, it means that the details will have to be computed every time. Although for small grammars this is fine, for large grammars this mean's a longer response time from the generated parser, and this can be a annoying thing to notice when your programming in an IDE/editor.

Like I said above, since the idea isn't bad, this feature should be an opt-in behaviour, as by default most developers wont be needing the comments:
```pegjs
MultiLineComment
  = "/*" comment:$(!"*/" SourceCharacter)* "*/" {
    return options.comments ? addComment(comment, true) : comment;
  }

MultiLineCommentNoLineTerminator
  = "/*" comment:$(!("*/" / LineTerminator) SourceCharacter)* "*/" {
    return options.comments ? addComment(comment, true) : comment;
  }

SingleLineComment
  = "//" comment:$(!LineTerminator SourceCharacter)* {
    return options.comments ? addComment(comment, false) : comment;
}
```

Mingun:
> I'm sorry but due to your choice of words I don't quite understand

Sorry. These are the autotranslator's expenses. I try to correct it according to my knowledge, but not always it turns out. I just trying to say, that solution must be flexible and powerfull and allow say "B" when it allow say "A". There is nothing worse when you understands what because of foolish restrictions suddenly appears it is impossible to implement that, it seemed - has to be implemented easily.

> As you can see from the source, `peg$computePosDetails` will pull the details from `peg$posDetailsCache` array, but only if details have been cached

Ah, this hashtable... Yes, now I understood.

> this feature should be an opt-in behaviour, 

Sounds reasonable. I update PR with option to enable comments extraction and also append README note. Spelling may be worse, so please, correct it if they need. I don't know English enough for writing correct texts.

Mingun:
Some comments?

futagoza:
Hi @Mingun, can you:

1) Remove the changelog entry. I'll add that my self when it's time to release 0.11
2) As this is an optional feature, using `varyParserOptions` in **test/unit/parser.spec.js** is overkill, as it is calling 48 tests 3 times in total. Only 1 test needs to be added, with the assertions inside confirming the result of a parse run with each of the options object similar to the ones in `optionsVariants`

Mingun:
If not to check all tests for all possible options of options, then how to guarantee that nothing breaks? I don't see any problems in executing 48 tests 3 times. Besides, I won't understand how you suggest to change a code and that it will test?

I remove chagelog entry. Also, the remark about changelog concerned only in `Changelog.md` or `Readme.md` too (I also want to pay attention that I have specially allowed editing pull request in order that you could correct the text in case of need)?

futagoza:
Most of the other 48 test's are testing some rule other then the `grammar` and don't even check the `comments` property in the root AST, so having them run multiple times is pointless. As this is an optional feature, the only tests needed are to check if it works and produces the correct result when `options.extractComments` is `true`.

> I remove chagelog entry. Also, the remark about changelog concerned only in Changelog.md or Readme.md too (I also want to pay attention that I have specially allowed editing pull request in order that you could correct the text in case of need)?

Yes, only the changelog entry needed to be removed, thank you. As for the `README.md`, I will correct any spelling and grammar error's if needed, before or after the merge.

Mingun:
Now I finish with this. Ready to merge

PR #513, opened 2017-07-08
Add JSON Schema for describing AST, generated by the parser

Mingun:
For authors of plug-ins it will be simpler to understand structure of AST and what opportunities are available for them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "babelify": "7.3.0",
     "browserify": "13.1.0",
     "chai": "3.5.0",
+    "chai-json-schema": "^1.4.0",
     "del": "2.2.2",
     "eslint-config-dmajda": "1.0.0",
     "express": "4.14.0",
--- a/src/ast.schema.json
+++ b/src/ast.schema.json
@@ -1,0 +1,276 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+
+  "type": "object",
+  "properties": {
+    "type": { "enum": ["grammar"] },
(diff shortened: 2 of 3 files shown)


futagoza:
Although this is a great idea @Mingun, I won't accept for 2 reasons:

1) This is something that should be released on NPM as a helper for plugin's that can be used with PEG.js and it's fork's, rather then a built in feature.
2) A `.d.ts` (#519) file is a better alternative that can be included with PEG.js

Mingun:
As far as I understand, d.ts files are necessary to the compiler TypeScript. It isn't absolutely clear to me how they will help JavaScript to developers or developers in other languages which will decide to work to generate/work with the existing AST.

Storage of the scheme separately from the main code can lead to desynchronizing it with the code that isn't really good while support of the scheme in the project is practically no-cost task.

futagoza:
> As far as I understand, d.ts files are necessary to the compiler TypeScript. It isn't absolutely clear to me how they will help JavaScript to developers or developers in other languages which will decide to work to generate/work with the existing AST.

`.d.ts` files can be used by VS Code, it's forks (as well as similar editors) and some libraries and tools (I'm currently working on a Game Editor that uses `.d.ts` files to verify the structure and types for AST returned from parsing JSON as well as a custom JavaScript based language, but it's a WIP tool).

> Storage of the scheme separately from the main code can lead to desynchronizing it with the code that isn't really good while support of the scheme in the project is practically no-cost task.

That won't happen if the developer (in this case I presume you) has a continued interest in using a JSON schema.

PR #528, opened 2017-09-03
Reimplement offset() and add range().

### LINKED ISSUE #526: Reimplement offset() helper in generated parsers
Can we get offset() back? Use case: a parser that builds a parse tree and annotates each node with its source location. offset() is very cheap, and calling it for thousands of nodes is fine. location() is much more expensive, and using it is a noticeable performance hit.

felix9:
offsetEnd() is new.

offset() existed prior to 0.10, and the existing name is why I'm not
calling it offsetStart().

Fixes #526


jaubourg:
You could name it `offsetStart` and alias it as `offset` for backward compatibility. Best of both worlds.

felix9:
Sure, but I lean slightly in favor of avoiding unnecessary redundancy. I'll add the alias if anyone feels strongly about it

futagoza:
@felix9 I was planning on adding `range()` instead of `offsetEnd()` as the reason most people look for the ending offset is when making a range:

```js
function range() {
    return [peg$savedPos, peg$currPos];
}
```

Any particular reason (and/or use case) you chose to use `offsetEnd()` instead?

felix9:
No reason. Range is probably better, yeah.

PR #529, opened 2017-09-08
Update package.json bin reference

tilgovi:
The bin field should reference the actual file. If it should be
installed as "peg" then it needs to mentioned as the "peg" key in a
"bin" mapping.
    
Additionally, the bin files need to be included in the list of published
files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "CHANGELOG.md",
     "LICENSE",
     "README.md",
-    "bin/pegjs",
+    "bin/peg.js",
+    "bin/options.js",
+    "bin/usage.txt",
     "examples/arithmetics.pegjs",
     "examples/css.pegjs",
     "examples/javascript.pegjs",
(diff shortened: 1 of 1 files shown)


tilgovi:
I'm fairly sure I got this right. I didn't try leaving off the '.js' to see if it's optional, but I'm guessing it is not. I'm assuming the move from pegjs -> peg was intentional, but I can change that if we want it to install as "pegjs".

futagoza:
Woop, my mistake, I forgot to update the files field 🤣 

The bin field doesn't need to be changed:
> If you have a single executable, and its name should be the name of the package, then you can just supply it as a string.
*https://docs.npmjs.com/files/package.json#bin*

I'm pretty sure this has been how it worked even before Node.js v4, so there's no need to add the extra bytes. Once you have reverted this change, I will squash and merge the changes to the files field

futagoza:
@tilgovi I've [pushed a commit](https://github.com/pegjs/pegjs/commit/14948dad9394c311b635bc52b2b9151e32c59e89) that updates the files field in the `package.json`, but the bin field is the same as that doesn't need changing.

tilgovi:
Thank you. Sorry I didn't follow up more quickly. I had a laptop meltdown this past week 💻 🔥 .

PR #532, opened 2017-09-12
Check allowedStartRules exist.

### LINKED ISSUE #524: Start rules passed in options are not checked
Currently the compiler doesn't check any rule passed to it via `options.allowedStartRules` so you can pass it anything and the parser is generated without any problems, but when you try to run the generated parser, the JavaScript engine will throw an error saying `xx is not defined`

I'm sure this wasn't an intended feature, so I'm labelling this as a `Bug`.

Will fix this before **0.11**, but PR's welcome 😄

felix9:
Fixes #524


futagoza:
🍪 🍖

PR #534, opened 2017-09-15
Improve error message for unbalanced brace.

felix9:
Currently, an open brace without a matching close brace will emit this confusing error message:

> Expected "!", "$", "&", "(", "*", "+", ".", "/", "/*", "//", ";", "?", character class, code block, comment, end of line, identifier, literal, or whitespace but "{" found.

This change adds an error case to the grammar to make it clearer what the problem is.


YemSalat:
@felix9 perhaps should add some tests for this?

PR #539, opened 2017-10-23
following es6 style

gsun:
revise the code following es6 style.


futagoza:
The minimum supported version of Node.js is v4, and while it has support for some ES6 features, default parameters is not supported till Node.js v6

PR #542, opened 2017-11-06
Replace custom `clone(...)` function with `Object.assign({}, ...)`

Mingun:
More standard way to clone objects

### DIFF
--- a/lib/compiler/passes/generate-bytecode.js
+++ b/lib/compiler/passes/generate-bytecode.js
@@ -202,16 +202,6 @@ function generateBytecode(ast) {
     );
   }
 
-  function cloneEnv(env) {
-    let clone = {};
-
-    Object.keys(env).forEach(name => {
-      clone[name] = env[name];
-    });
-
-    return clone;
-  }
-
   function buildSequence() {
     return Array.prototype.concat.apply([], arguments);
   }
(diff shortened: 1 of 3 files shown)


futagoza:
It is a more standard way to clone objects and also make options with defaults, but it's also an ES6 built-in function, therefore won't be available in some environments that are targeted by PEG.js

I'm working on adding a util module that will have these small helpers that are spread through the PEG.js codebase, similar to how PEG.js had one in the pre-0.10 codebase, but this comes after I'm done with and merged the [development-overhaul branch](https://github.com/pegjs/pegjs/tree/development-overhaul) into the master branch.

Mingun:
> therefore won't be available in some environments that are targeted by PEG.js

?!  the pegjs code itself already use ES6 features, so I misunderstand how this function may be not availible. This PR not affect generated parsers, that still generated for ES5 environment.

futagoza:
Yes, it does, but only ES6 syntax that can be transpiled down to ES5, but I prefer not to add ES6 built-ins as Babel (used via Browseriefy in the PEG.js codebase, but in general also) and it's plugins tend to add extra weight for some polyfills as they follow the standard.

If I can, I try to avoid polyfills, although in some cases it can't be avoided. In this case, it can, so sorry 😄

PR #547, opened 2017-12-20
Partial fix #194 - nonsenses error messages with semantic predicates

### LINKED ISSUE #194: Syntactic predicates cause nonsensical error messages
With the following grammar (that matches stuff like "abc", "aabbcc", "aaabbbccc"...), empty test code shows 'Line 1, column 1: Expected end of input but end of input found'. I think it should say 'Expected "a"' instead, as lookahead is looking for 'a'.

```
S = &(A !'b') 'a'+ B
A = 'a' A? 'b'
B = 'b' B? 'c'
```

Mingun:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Absolutely correct fix is very difficult, and shall include uncommon methods of merge of the expected messages for compilation of lungs for perception of error messages. However, as still did not complain of errors (except this issue), implementation of complete fix can be postponed.

This fix introduces negative expectations and also expectations from positive lookaheads. For grammar
```pegjs
start = !'a';
```
and input `a` error will be
```
Expected not "a", but "a" found
```
This is still slightly incorrect, because if you type character, other than "a", then there will be other parse error, but it nevertheless is better, than being earlier
```
Expected , but "a" found
```
Ideally such grammar must not compile at all, because it have contradictory requirements: on the same place at the same time are expected a character "a" and EOF. I have prototype, but they will be ready not soon.

### DIFF
--- a/test/spec/behavior/generated-parser-behavior.spec.js
+++ b/test/spec/behavior/generated-parser-behavior.spec.js
@@ -1194,14 +1194,28 @@ describe( "generated parser behavior", function () {
 
                 } );
 
-                it( "discards any expectations recorded when matching the expression", function () {
+                it( "doesn't discard any expectations recorded when matching the expression", function () {
 
                     const parser = peg.generate( "start = 'a' / &'b' / 'c'", options );
 
                     expect( parser ).to.failToParse( "d", {
                         expected: [
                             { type: "literal", text: "a", ignoreCase: false },
+                            { type: "literal", text: "b", ignoreCase: false },
                             { type: "literal", text: "c", ignoreCase: false }
+                        ]
+                    } );
+
+                } );
+
+                it( "records expectations from right place", function () {
+
+                    const parser = peg.generate( "start = 'a' / &'b' .", options );
+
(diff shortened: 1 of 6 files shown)


test/spec/behavior/generated-parser-behavior.spec.js
```diff
@@ -1194,19 +1194,33 @@ describe( "generated parser behavior", function () {
 
                 } );
 
-                it( "discards any expectations recorded when matching the expression", function () {
+                it( "not discards any expectations recorded when matching the expression", function () {
```
futagoza: this would make more sense: "doesn't discard any expectations recorded when matching the expression"

Mingun:
This ready to merge

Issue #517, opened 2017-07-20
Using external functions in parser actions


stuhlmueller:
I'm using pegjs in node and would like to use external functions, classes, and libraries in parser actions. Here's a minimal example:

```js
import peg from "pegjs";

class Wrapper {
  constructor(text) {
    this.text = text;
  }
}

var parser = peg.generate("start = ('a' / 'b')+ { return new Wrapper(text); }");

parser.parse("abba");
```

Of course, as is this fails with `ReferenceError: Wrapper is not defined`. Is there a way to make this work?

jtenner:
You could try:

```javascript
//./wrapper.js
class Wrapper {
  constructor(text) {
    this.text = text;
  }
}

import peg from "pegjs";

var parser = peg.generate(`
{ let Wrapper = require('./wrapper'); }

start = ('a' / 'b')+ { return Wrapper(text); }
`);

parser.parse("abba");
```

stuhlmueller:
This is a promising direction and works for absolute imports.

Relative imports fail though:

    Error: Cannot find module './wrapper'.

I think the reason is that node is searching in `pegjs/lib/compiler` (within `node_modules`), not in the main script directory. It's possible to construct a relative path from there to `wrapper.js`, e.g. `../../../src/wrapper.js`, and this works, but I wonder if there's a better approach.

Mingun:
Inside the parser you have access to `options`  variable, so just use that code:
```js
let peg = require("pegjs");
 
let parser = peg.generate(
  "start = ('a' / 'b')+ { return new options.Wrapper(text); }",
  { Wrapper: Wrapper }
);
 
 parser.parse("abba");
```

stuhlmueller:
Neat! Exactly what I was looking for:

```js
const parser = peg.generate(`
{ const { Wrapper } = options; }
start = ('a' / 'b')+ { return new Wrapper(text); }
`);

parser.parse("abba", { Wrapper });
```

Izhaki:
@Mingun Your example doesn't seem to work (providing options in `generate`) whereas @stuhlmueller example does (providing options in `parse`).

I'd argue that it would be nice to be able to pass options to `generate`, because often what you want is provide support functions to the grammar, and not something that is parse specific. Would be nice if the second parameter given to `generate` will also have an `helpers` field that you can access in the parser.

[...]

PR #552, opened 2018-01-06
Improve error when reserved word used as label

Mingun:
Before this commit error looks like (for input `start = break:'a'`)

> Expected "!", "$", "&", "(", "*", "+", ".", "/", "/*", "//", ";", "?", character class, code block, comment, end of line, identifier, literal, or whitespace but ":" found.

After this error looks like

> Expected identifier but reserved word "break" found.

(cherry picked from commit 9c8b868f406dc50f1a15690c9938d90034f6025d)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -198,9 +198,13 @@ function peg$parse(input, options) {
   var peg$c10 = ":";
   var peg$c11 = peg$literalExpectation(":", false);
   var peg$c12 = function(label, expression) {
+        if (RESERVED_WORDS.indexOf(label[0]) >= 0) {
+          error(`Label can't be a reserved word "${label[0]}".`, label[1]);
+        }
+
         return {
           type: "labeled",
-          label: label,
+          label: label[0],
           expression: expression,
           location: location()
         };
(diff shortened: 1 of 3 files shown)


futagoza:
In `src/parser.pegjs` can you:

- rename `RESERVED_WORDS_JS` to `RESERVED_WORDS`
- remove `let reservedWords = new Set(RESERVED_WORDS_JS);` in the initializer block
- change the code in the action block for `Identifier` to
```js
if ( RESERVED_WORDS.indexOf(name) === -1 ) return name;
error(`Expected identifier but reserved word "${name}" found.`);
```

The reason for these changes is simple:

1. It is already obvious what these are reserved for, so no need to add the ugly `_JS` postfix
2. Using `Set` will make Babel add a polyfill for it, and we only need to check if `name` exists within the list, so a simple use of `Array#indexOf` is better here.

Mingun:
> It is already obvious what these are reserved for, so no need to add the ugly `_JS` postfix

Potentially the code can be generated not only for JS and reserved keywords will be others. Originally I wanted to add a feature to allow to redefine in options this list (for support of other languages), but now there is no opportunity to supply options which will be used only in compilation stage (#517).

Mingun:
This can't be merged in such way, as it broke, for example, JSON example grammar. `error` and `expected` functions caused immediately reporting error instead of backtraking, and when grammar contains rule with reserved name it dailed to parse, because it firstly try parse `label:experession` and then `expression`/ Example of such grammar:
```
start = false;
false = 'a';
```

Mingun:
But now this problem fixed.

futagoza:
> Potentially the code can be generated not only for JS and reserved keywords will be others. Originally I wanted to add a feature to allow to redefine in options this list (for support of other languages), but now there is no opportunity to supply options which will be used only in compilation stage (#517).

I've taken note of this (facepalmed my self, of course, wondering why I forgot this when I did the same hack my self a while back 😆 ) and will add this after I'm done with the ASTVisitor class.

PR #568, opened 2018-03-15
add missing eslint config file as referenced by rc

StoneCypher:
eslint needs a private config `eslint-config-futagozaryuu` from which to pull `eslint-config-futagozaryuu/node-v4`.  probably missing because, as is a personal config, it's likely installed globally on the owner's machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
         "coveralls": "3.0.0",
         "dedent": "0.7.0",
         "del": "3.0.0",
-        "eslint-config-futagozaryuu": "3.9.x",
+        "eslint-config-futagozaryuu": "^3.9.0",
         "express": "4.16.2",
         "glob": "7.1.2",
         "gulp": "3.9.1",


futagoza:
Actually, `eslint-config-futagozaryuu` is published on NPM, and should install fine on any machine other then mine (for example, travis tests pass even with this pr).

What error or problem did you encounter exactly to submit this PR?

StoneCypher:
eslint wouldn't lint the project because of the missing module

futagoza:
Details about your setup might help. I've used this config while working on:

- Node.js v6 (Windows 10 x32 and Windows 10 x64)
- Node.js v7 and v8 (Windows 10 x64)
- Node.js v4, v5, v6, v7 and v8 (Linux, Travis)

While using it on the above, I have not had this problem on PEG.js, but I have had the problem quite a while back on another project (haven't put it on GitHub) where it couldn't find my ESLint configurations, but after clearing my NPM cache and deleting the local copy of node_modules it was fixed, so maybe something similar could help you?

Also to note, although I don't use carets much, I'm pretty sure there's no difference between `3.9.x` and `^3.9.0`, as `^3.9.0` means any release between `3.9.0` and `3.10.0`, which is the same as `3.9.x`

futagoza:
@StoneCypher I closed this because

1. I am unable to reproduce the issue (on either of my laptops, or on Travis)
2. The PR change `^3.9.0` is the same as the orignal `3.9.x`
3. I updated `eslint-config-futagozaryuu` to `4.17.x`

From what I can tell, you are probally having an issue with dependency insallation, so try deleting the `node_modules` folder on your fork of the PEG.js repository, as well as clearing your npm cache, then do a fresh `npm install` from the root of your fork'.

If you are still having issues, or you beleive that I am wrong, then please open a new issue and tell me, because if this is an issue resulting from my choice of `x` instead of carets (`^`) in a version string, then I want to fix it, as I use `x` in nearly all the packages or projects I develope.

PR #608, opened 2019-04-14
Add location to buildMessage

vetedde:
It helps create usefull message with postioin of error

### PR type

Add new parameter location to buildMessage function

### Prerequisites

To make the bug messages more readable I needed to know their location. For this, I pass as an additional parameter location to buildMessage function/

- **I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document:**
- **I have updated the documentation accordingly:**
- **I have added tests to cover my changes:**

### Description

P.S. It is my first pull request, I tried to do everything right, but if I made a mistake, I’m ready to fix it if this pull seems good, but unfinished


Mingun:
You add a new parameter but do not use it?..

vetedde:
> You add a new parameter but do not use it?..

No, I use it, when I customize my error message

'''
SyntaxError.buildMessage = function(expected, found, location) {
 return `Excpected ${describeExpected(expected)}, but found ${found} at ${ location.start.line }`;
}
'''

Mingun:
There are some reasons for which it is necessary to replace `buildMessage` to change the message? In thrown exception you have all information to make any message what you want

vetedde:
There is a built-in ability to customize the message that will be displayed to the user with help buildMessage. I don't see the point of writing error handling to display the user the same message, but with the location, so I just added a location parameter

futagoza:
@Mingun I think this is fine, as some people don't like using `try...catch`, and overriding the default behaviour of this function gives them the ability to customise the error messages without using `try...catch`. This PR simply gives them the location as well.

I'm fine with merging this once you:

- There's no need to commit these files because of a removed empty line:
  - packages/pegjs/lib/peg.js
  - packages/pegjs/bin/peg.js
- Remove all _package-lock.json_ files
- Don't edit `website/vendor/pegjs/peg.js`, thats PEG.js v0.10

PR #622, opened 2019-07-31
Adds support for arrow function: #580

### LINKED ISSUE #580: Shorthand for Semantic Actions
It would be nice to add shorthand for semantic actions.

Say, instead of writing `{ return value }`, we write, for example `{ extract }` which is defined in the initializer.

For example:
```js
{
  function extract(value) {
    return value;
  }
  function concat(head, tail) {
    return head ? [head].concat(tail) : [];
  }
  function toAddExpr(head, tail) {
    return { type: 'addExpr', expressions: concat(head, tail) };
  }
}

List
  = '(' _ head:Item? tail:( _ ',' _ value:Item { extract } )* _ ')' { concat }

// Another kind of list
Add
  = '(' _ head:Multiply? tail:( _ '+' _ value:Multiply { extract } )* _ ')' { toAddExpr }
```

First, this will make us able to reuse functions... in a nicer way :smile: 
Second, this will make our grammar a bit more readable.

I believe it also will be more useful when the expression contained in the action shorthand is a member expression `{ foo.bar.baz }` instead of just identifier `{ foo }`. So that grammar writers can organize their functions inside of an object or even a module.

proteriax:
### PR type

- **New feature (non-breaking change which adds functionality):**

### Prerequisites

- **I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document:** yes
- **I have updated the documentation accordingly:** on it
- **I have added tests to cover my changes:** yes

### Description

This pull requests adds the following shorthand grammar:

```
foo
  = '(' expr:some_expression ')' => expr
  ;
```


polkovnikov-ph:
I'm afraid `code:Code` clause is way too permissive. It should match to EcmaScript expression, not arbitrary piece of ES code.

proteria:
The current `Code` block does not validate the internal content either so you can put complete gibberish inside and it still produces a parser. Sure you can include a JavaScript AST parser like `acorn`, but that would be a huge breaking change and get into the mess that is allowing JavaScript parser options and experimental syntax supports.

polkovnikov-ph:
I was worried that it's not confined by anything but the end of the line. Is it intended behavior that the rest of the line is considered a `Code`?

futagoza:
@polkovnikov-ph is right, using just `EOS` will stop this from being used in cases like this:

```js
{ const blacklisted = [ "Hi", "You" ]; }

Greeting
  = "Hi" " "+ ( name:$[a-z\-]i+ !=> blacklisted.includes( name ) ) => text()
```

For now, replacing `EOS` with `(_ / EOS)` should allow cases like that, and then later this can be solved by [implementing common ECMAScript related expressions after fat arrows](https://github.com/pegjs/pegjs/issues/580#issuecomment-435282004)

Issue #586, opened 2018-10-22
Full Unicode support, namely for codepoints outside the BMP


doersino:
### Issue type

- **Bug Report:** yes
- **Feature Request:** kinda
- **Question:** no
- **Not an issue:** no

### Prerequisites

- **Can you reproduce the issue?:** yes
- **Did you search the repository issues?:** yes
- **Did you check the forums?:** yes
- **Did you perform a web search (google, yahoo, etc)?:** yes

### Description

JavaScript is, without [some custom boilerplate](https://stackoverflow.com/questions/3744721/javascript-strings-outside-of-the-bmp), unable to properly deal with Unicode characters/codepoints outside the [BMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane), i.e., ones whose encoding requires more than 16 bits.

This limitation seems to carry over to PEG.js, as shown in the example below.

In particular, I'd like be be able to specify ranges such as `[\u1D400-\u1D7FF]` (which presently turns into `[ᵀ0-ᵁ9]`) or equivalently `[𝐀-𝐙]` (which throws an "Invalid character range" error).

(Might there be a way to make this work that does not require changes to PEG.js?)

### Steps to Reproduce

1. Generate a parser from the grammar given below.
2. Use it to try to parse something ostensibly-conforming.

**Example code:**

This grammar:

```js
//MathUpper = [𝐀-𝐙]+
MathUpperEscaped = [\u1D400-\u1D7FF]+
```

**Expected behavior:**

The parser generated from the given grammar successfully parses, for example, "𝐀𝐁𝐂".

**Actual behavior:**

A parse error: `Line 1, column 1: Expected [ᵀ0-ᵿF] but "` (Or, when uncommenting the other rule, an "Invalid character range" error.)

### Software

- **PEG.js:** 0.10.0
- **Node.js:** Not applicable.
- **NPM or Yarn:** Not applicable.
- **Browser:** All browsers I've tested.
- **OS:** macOS Mojave.
- **Editor:** All editors I've tested.

futagoza:
[...]

PR #651, opened 2020-06-04
Use ES6 syntax for astral Unicode characters

Seb35:
### PR type

- **Bug fix (non-breaking change which fixes an issue):** _no_
- **New feature (non-breaking change which adds functionality):** _yes_
- **Breaking change (fix or feature that would cause existing functionality to change):** _no_
- **Documentation change:** _no_

### Prerequisites

- **I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document:** _yes_
- **I have updated the documentation accordingly:** _no_ but it should be done indeed
- **I have added tests to cover my changes:** there is one

### Description

This PR is constructed on #586 but uses the ES 6 syntax for astral Unicode characters `\u{1F4AF}` instead of the syntax `\U0001F4AF` (recognised in Python world).

The rule `CodePoint` (named after [ECMAScript grammar](http://www.ecma-international.org/ecma-262/10.0/#prod-CodePoint)) recognises the following sequences:
* `\u{0+}` e.g. `\u{00000000000000}` resulting in the Unicode character U+0000
* `\u{0*HexDigit{1,4}}` resulting in the Unicode character in BMP
* `\u{0*(10|HexDigit)HexDigit{4}}` resulting in the Unicode character in astral plane

This PR or #586 have a similar goal but a different proposed syntax.

They are both limited to constant strings and do both a bad job in characters classes like `[a-\u{1F4AF}]`. Characters classes should be treated by further PRs, see [my comment](https://github.com/pegjs/pegjs/issues/586#issuecomment-638867416) for other points.


hildjj:
Can you port this over to merge cleanly on peggyjs/peggy please?  I'd like this to be in the first peggy release.

hildjj:
Actually, not before peggy 1.0.  See: https://github.com/peggyjs/peggy/discussions/15

